PR #4, opened 2023-01-30
docs: initial API spec draft

iloveagent57:
**Description:**
Defines a formal API spec for the subsidy application.
Note some remaining open questions in the first chapter of the doc.

**JIRA:**
https://2u-internal.atlassian.net/browse/ENT-6741

**Reviewers:**
- [ ] @pwnage101 
- [ ] @johnnagro 

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Commits are squashed

### DIFF
--- a/docs/decisions/0002-initial-api-specification.rst
+++ b/docs/decisions/0002-initial-api-specification.rst
@@ -1,0 +1,448 @@
+0002 Initial API Specification
+##############################
+
+Status
+******
+
+**Provisional**
+
+Context
+*******
+
+This doc describes the API specification for the ``enterprise-subsidy`` service.
+Consider each chapter below a "decision".
+
+Notable Implementation Details
+==============================
+
+- A reference to the policy that allowed a ledger transaction to be created
+  should be written on the ledger transaction record.
+  This is important in the context of a learner-credit access policy that
+  has a spending cap imposed for the whole policy.
+- The enterprise-subsidy service should always ask course-discovery for the current price of a piece of content
+  before creating a ledger transaction, or when answering if a query about whether a ledger transaction
+  can be written (i.e. "redeemability") for a given user/content combination. The course-discovery service is the content
+  metadata source of truth and has certain compliance obligations, as does the enterprise-subsidy service.
+  The enterprise-subsidy service may cache content price (or other metadata) for up to five minutes, as we expect
+  that clients may use the service in such a way that 2 or 3 subsequent requests for price data may be made
+  in a short amount of time. This should limit additional load placed on the course-discovery service from
+  the enterprise-subsidy service.
+- As a client, the access policy API can expect that the enterprise-subsidy service will answer
+  redeemability queries with both a boolean answer *and* the current price of the content in question.
+
+Open Questions and TODOs
+========================
+
+- We should formally declare "nothing about groups needs to be modeled at this time" for both the Subsidy API
+  and its primary client, Subsidy Access Policy API.
+
+Primary Subsidy Route
+*********************
+**/api/v1/subsidies/**
+The root URL for reading simple metadata about enterprise subsidies.
+
+GET (list) enterprise subsidies
+===============================
+**/api/v1/subsidies/?enterprise_customer_uuid=[customer-uuid]&subsidy_type={learner_credit,subscription}**
+
+Inputs
+------
+
+- ``enterprise_customer_uuid`` (query param, optional): Specifies the enterprise for which associated subsidy metadata records should be returned.
+- ``subsidy_type`` (query param, optional): Specifies what type of subsidy metadata records should be returned.  Allowed values are ``learner_credit`` and ``subscription``.
+
+Outputs
+-------
+Returns a paginated list of subsidy metadata records of the form:
+
+::
+
+   {
+       'count': 1,
+       'next': null,
+       'previous': null,
+       'results': [
+           {
+               'uuid': 'the-subsidy-uuid',
+               'enterprise_customer_uuid': 'the-enterprise-uuid',
+               'active_datetime': '2023-01-01T00:00:00Z',
+               'expiration_datetime': '2024-01-01T00:00:00Z',
+               'title': 'The Learner Credit Subsidy for my Enterprise',
+               'subsidy_type': 'learner_credit',
+               'unit': 'USD_CENTS',
+               'opportunity_id': 'some-opp-id',
+               'remaining_balance': 987650
+           }
+       ]
+   }
+
+For some types of subsidies, e.g. subscriptions, this payload may also include things like `subscription_plan_uuid`.
+
+Permissions
+-----------
+
+**enterprise_admin**
+  Should only list subsidies for which the requesting user has implicit (JWT) or explicit (DB-defined) access.
+  Optionally filtered to only the enterprises specified by the ``enterprise_customer_uuid`` query parameter.
+
+**openedx_operator**
+  If the requesting user is implicitly or explicitly granted this role, they have permissions to view **all**
+  subsidy records.  Optionally filtered to only the enterprises specified by the ``enterprise_customer_uuid`` query parameter.
+
+
+GET (retrieve) enterprise subsidy
+=================================
+**/api/v1/subsidies/[subsidy-uuid]/**
+
+Inputs
+------
+
+- ``subsidy-uuid`` (URL path, required): The uuid (primary key) of the subsidy to retrieve.
+
+Outputs
+-------
+Returns a single subsidy metadata records of the form:
+
+::
+
+   {
+       'uuid': 'the-subsidy-uuid',
+       'enterprise_customer_uuid': 'the-enterprise-uuid',
+       'active_datetime': '2023-01-01T00:00:00Z',
+       'expiration_datetime': '2024-01-01T00:00:00Z',
+       'title': 'The Learner Credit Subsidy for my Enterprise',
+       'subsidy_type': 'learner_credit',
+       'unit': 'USD_CENTS',
+       'opportunity_id': 'some-opp-id',
+       'remaining_balance': 987650
+   }
+
+Permissions
+-----------
+
+enterprise_admin
+  Should return the requested subsidy only if the requesting user has implicit (JWT) or explicit (DB-defined)
+  ``enterprise_admin`` role assigned for the requested subsidy's enterprise.
+
+openedx_operator
+  If the requesting user is implicitly or explicitly granted this role, they have permissions to view **all**
+  subsidy records, and can therefore retrieve the requested subsidy.
+
+Disallowed actions/verbs
+========================
+None of ``POST``, ``PATCH``, ``PUT``, or ``DELETE`` should be supported at this time.
+Currently, subsidy records must be created by a staff user via Django Admin. We don't support the creation
+of subsidies via a user-facing UI or from any other service.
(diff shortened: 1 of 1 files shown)


docs/decisions/0002-initial-api-specification.rst
```diff
@@ -0,0 +1,436 @@
...
+
+::
+
+   {
+       'previous': null,
+       'next': '...',
+       'count': 'blah',
+       'aggregates': {
+           'total_quantity': 12350,
+           'unit': 'USD_CENTS',
```
pwnage101, line 171: What if the resulting transactions contain a mix of units?  Should we somehow design for that eventuality, or just commit to only aggregating USD transactions for now?

docs/decisions/0002-initial-api-specification.rst
```diff
@@ -0,0 +1,436 @@
...
+               'uuid': 'the-transaction-uuid',
+               'status': 'completed',  TODO: enumerate valid statuses
+               'idempotency_key': 'the-idempotency-key',
+               'user_id': 54321,
+               'content_key': 'demox_1234+2T2023',
+               'quantity': 19900,
+               'unit': 'USD_CENTS',
+               'reference_id': 1234,
+               'reference_table': 'enrollments',
+               'subsidy_access_policy_uuid': 'a-policy-uuid',
```
pwnage101, line 185: Note to self that I think this is not tracked at the Transaction model level and we should probably add a field for it.  Assuming this is the policy that was "used" to create the redemption/transaction.

docs/decisions/0002-initial-api-specification.rst
```diff
@@ -0,0 +1,436 @@
...
+               'metadata': {...},
+               'created': 'created-datetime',
+               'modified': 'modified-datetime',
+               'reversal': null
```
pwnage101, line 189: can `reversal` be a list?  The model relationships aren't currently enforcing a 1:1 relationship (which I'm guessing is to support multiple partial reimbursements in the future).

docs/decisions/0002-initial-api-specification.rst
```diff
@@ -0,0 +1,436 @@
...
+Permissions
+-----------
+
+enterprise_learner
+  The transaction data returned should be filtered to only those related to the requesting user, within
+  the subsidy specified in the request's URL path.
+
+enterprise_admin
+  Should return the requested subsidy transactions only if the requesting user has implicit (JWT) or explicit (DB-defined)
+  ``enterprise_admin`` role assigned for the requested subsidy's enterprise.
```
pwnage101, line 203: Curious in general about these non-operator roles across all the subsidy/ledger endpoints: will we ever expect that a customer or learner will interact directly with these endpoints?  My understanding was that access would either be granted to ECS/engineering (openedx_operator), or the backend service enterprise-access (also openedx_operator?).

docs/decisions/0002-initial-api-specification.rst
```diff
@@ -0,0 +1,436 @@
...
+enterprise_admin
+  NOT ALLOWED
+
+openedx_operator
+  If the requesting user is implicitly or explicitly granted this role, they have permissions to reverse
+  transactions for any active ledgered-subsidy.
+
+POST can redeem in subsidy
+==========================
+**/api/v1/subsidies/[subsidy-uuid]/can_redeem/**
```
pwnage101, line 374: does this POST endpoint have a side-effect?

docs/decisions/0002-initial-api-specification.rst
```diff
@@ -0,0 +1,436 @@
...
+- Subsidies and transactions only relate to a content `catalog` via a reference to the `subsidy access policy`
+  which caused a transaction to be created.  We assume that whatever (user, content key) combination is contained
+  in the payload to the ``POST /api/v1/subsidies/.../transactions/`` endpoint has already been verified for inclusion
+  in an appropriate catalog.  This is unconcerning at present, as we will only allow service-users or openedx operators
+  to create transactions.
+
+Rejected Alternatives
+*********************
+
+- A more fully-fledged API.  We'll work toward this in the future, but are starting with the bare minimum for now.
```
pwnage101, line 426: haha, a broadly better solution REJECTED

docs/decisions/0002-initial-api-specification.rst
macdiesel: For now we should just commit to USD. What I think is likely to happen is that all transactions will be stored in a singular unit type and then when the combination of metadata about the transaction, contract data and historical currency exchange rates we would reconstruct the ledger in a different unit and then do the aggregation.

docs/decisions/0002-initial-api-specification.rst
```diff
@@ -0,0 +1,436 @@
...
+       'results': [
+           {
+               'uuid': 'the-transaction-uuid',
+               'status': 'completed',  TODO: enumerate valid statuses
+               'idempotency_key': 'the-idempotency-key',
+               'user_id': 54321,
+               'content_key': 'demox_1234+2T2023',
+               'quantity': 19900,
+               'unit': 'USD_CENTS',
+               'reference_id': 1234,
```
macdiesel, line 183: Do piggyback on what DE/BI mentioned earlier I think that overloading these fields will make the analysis harder in the long run. I think it would be better to have sparse data than to look up data by type.

docs/decisions/0002-initial-api-specification.rst
```diff
@@ -0,0 +1,436 @@
...
+
+openedx_operator
+  If the requesting user is implicitly or explicitly granted this role, they have permissions to view **all**
+  subsidy records, and can therefore retrieve the requested subsidy transactions.
+
+GET (retrieve) enterprise subsidy transaction
+=============================================
+**/api/v1/subsidies/[subsidy-uuid]/transactions/[transaction-uuid]/**
+
+This endpoint can retrieve a single transaction, provided the caller has the subsidy and transaction uuid.
```
macdiesel, line 213: Why do we need both the subsidy and transaction ID? Would it simplify this to just use the transaction ID?

docs/decisions/0002-initial-api-specification.rst
```diff
@@ -0,0 +1,436 @@
...
+  NOT ALLOWED
+
+enterprise_admin
+  NOT ALLOWED
+
+openedx_operator
+  If the requesting user is implicitly or explicitly granted this role, they have permissions to create
+  transactions for any active ledgered-subsidy.
+
+POST enterprise subsidy transaction reversal
```
macdiesel, line 314: I get why it's like this but for the record I like it semantically better if this was a delete on the URL previous to this but I get why you specified it this way.

docs/decisions/0002-initial-api-specification.rst
macdiesel: yeah... Should this be a GET?

macdiesel (commented):
Looks good Alex.

docs/decisions/0002-initial-api-specification.rst
pwnage101: for the curation app, we opted to use `/api/v1/hightlight-sets` instead of `/api/v1/enterprise-curations/[enterprise_uuid]/highlight-sets`, presumably because the highlight set UUIDs were already globally unique.  In the case of subsidies and transactions, I think the situation is probably the same, so I'd tentatively agree with beggs here.

docs/decisions/0002-initial-api-specification.rst
```diff
@@ -0,0 +1,436 @@
...
+Context
+*******
+
+This doc describes the API specification for the ``enterprise-subsidy`` service.
+Consider each chapter below a "decision".
+
+Notable Implementation Details
+==============================
+
+- A reference to the policy that allowed a ledger transaction to be created
```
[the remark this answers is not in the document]
pwnage101: Up to this point we've been talking about only allowing one policy to be used to create a transaction.  The idea being that the upstream subsidy-access-policy app will establish the list of applicable policies and then pick (resolve) only one to use.

That said, If multiple applicable policies referred to the same subsidy, then who's to say which policy was used?

docs/decisions/0002-initial-api-specification.rst
pwnage101: I just noticed there's both a "total_quantity" and also a "remaining_balance".  Maybe that already answers my question, in that the former refers to seats and the latter refers to USD cents?

docs/decisions/0002-initial-api-specification.rst
```diff
@@ -0,0 +1,436 @@
...
+  should be written on the ledger transaction record.
+  This is important in the context of a learner-credit access policy that
+  has a spending cap imposed for the whole policy.
+- The enterprise-subsidy service should always ask course-discovery for the current price of a piece of content
+  before creating a ledger transaction, or when answering if a query about whether a ledger transaction
+  can be written (i.e. "redeemability") for a given user/content combination. The course-discovery service is the content
+  metadata source of truth and has certain compliance obligations, as does the enterprise-subsidy service.
+  The enterprise-subsidy service may cache content price (or other metadata) for up to five minutes, as we expect
```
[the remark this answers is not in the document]
iloveagent57: Will add a `Consequences` note about this.

docs/decisions/0002-initial-api-specification.rst
```diff
@@ -0,0 +1,436 @@
...
+
+Outputs
+-------
+Returns a single subsidy metadata records of the form:
+
+::
+
+   {
+       'uuid': 'the-subsidy-uuid',
+       'customer_uuid': 'the-enterprise-uuid',
```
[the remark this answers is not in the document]
iloveagent57: I've switched to `enterprise_customer_uuid` everywhere.

docs/decisions/0002-initial-api-specification.rst
```diff
@@ -0,0 +1,436 @@
...
+Inputs
+------
+
+- ``subsidy-uuid`` (URL path, required): The uuid (primary key) of the subsidy for which transactions should be listed.
+- ``include_aggregates`` (query param, optional): Specifies if aggregates (quantities, number of transactions) should be
+  returned as part of the paginated response.  Defaults to ``true``.
+- ``user_id`` (query param, optional): If present, filters returned transactions and/or aggregates to only
+  those associated with the specified ``user_id`` value.
+- ``content_key`` (query param, optional): If present, filters returned transactions and/or aggregates to only
+  those associated with the specified ``content_key`` value.
```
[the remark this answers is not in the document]
iloveagent57: Yes, we should switch to `content_key` everywhere, that's the canonical reference across services.

docs/decisions/0002-initial-api-specification.rst
iloveagent57: We should have a separate ADR about "double-entry" reference tracking across services.  I'll add a ticket to our backlog.

docs/decisions/0002-initial-api-specification.rst
iloveagent57: I've tried to structure it so that non-operators/non-staff/non-services can only read relevant data.  Will add a note in `Consequences`.

docs/decisions/0002-initial-api-specification.rst
iloveagent57: that's a fair point.  will change!

PR #7, opened 2023-02-08
Fix broken migration

pwnage101:
This PR first reverts and then reapplies the commit which included the bad migration, but without any of the model changes.  Then, it incrementally re-introduces the model changes, creating separate commits and separate migrations along the way.  This seems to solve the compatibility problem with sqlite.

The last migration `0005_add_subsidy_fields_20230208_0104.py` will fail in stage because the changes in that migration have already been completely applied.  The error message will be `Duplicate column name 'title'`.  Nevertheless, we should deploy this commit to stage, allow the last migration to fail, then have SRE manually fake the last migration:

```
python manage.py migrate --fake subsidy 0005_add_subsidy_fields_20230208_0104
python manage.py migrate
```

Rejected Alternatives:
---

One idea I had was to create a custom 0005 migration which was insensitive to the pre-existence of the title column.  Essentially I would extend AddField to add a "skip" branch based on the existence of the database column.  However, in the process of writing that, I realized it has at least two problems:

* It needs to support both MySQL *and* sqlite, which adds complexity.
* The logic depends on non-public methods in django core which are subject to change.


macdiesel (approved):
Once we get this in to prod I'm wondering if we should perhaps squash the migrations to get rid of this operational artifact?

iloveagent57 (approved):

PR #12, opened 2023-02-17
feat: basic rbac support

pwnage101:
edx-rbac support has been added, and rules defined to grant the following permissions for now:

* `subsidy.can_read_subsidies`
* `subsidy.can_read_transactions`
* `subsidy.can_create_transactions`

As we add support for more CRUD endpoints, more permissions may need to be created.  Barebones viewsets are provided, covering ADR 0002-initial-api-specification, and mapped to the permissions.

This commit also discards the SubscriptionSubsidy model for now, committing to a single Subsidy model for tracking learner credit.  An ADR has been added to this effect.

4 migrations are included:

* 0008 - [edx-rbac] supportng models for edx-rbac.
* 0009 - [subsidy models] Drop tables (deleting data) to prepare for migration to a single Subsidy model.
* 0010 - [subsidy models] Recreate single Subsidy model.
* 0011 - [edx-rbac] data migration to seed feature roles.

Note that 0009 is not functionally reversible, dropped tables are not recoverable.

ENT-6620


enterprise_subsidy/apps/subsidy/constants.py
```diff
@@ -0,0 +1,24 @@
+"""
+Constants for the subsidy app.
+"""
+
+ACCESS_TO_ALL_ENTERPRISES_TOKEN = '*'
```
iloveagent57, line 5: nit: this is defined by `edx_rbac.utils.ALL_ACCESS_CONTEXT`

enterprise_subsidy/apps/subsidy/rules.py
```diff
@@ -0,0 +1,140 @@
...
+        bool: True if the user has access.
+    """
+    if not context:
+        return False
+    return user_has_access_via_database(
+        user,
+        ENTERPRISE_SUBSIDY_OPERATOR_ROLE,
+        EnterpriseSubsidyRoleAssignment,
+        context,
+    )
```
iloveagent57, line 51: Nit: it's never struck me before how similar the implicit/explicit bodies of all our rules predicates are.  Might be worth adding a little private helper e.g.
```python
def _user_has_implicit_access_via_role(user, context, role):
    if not context:
        return False
    request = crum.get_current_request()
    decoded_jwt = get_decoded_jwt(request) or get_decoded_jwt_from_auth(request)
    return request_user_has_implicit_access_via_jwt(decoded_jwt, role, context)
```
Longer-term, we should add these helpers to `edx_rbac.utils`

enterprise_subsidy/apps/subsidy/rules.py
```diff
@@ -0,0 +1,140 @@
...
+    has_implicit_access_to_subsidy_admin | has_explicit_access_to_subsidy_admin
+)
+# pylint: disable=unsupported-binary-operation
+has_operator_level_access = has_implicit_access_to_subsidy_learner | has_explicit_access_to_subsidy_learner
+
+
+# Finally, grant specific permissions to the appropriate access level.
+rules.add_perm(PERMISSION_CAN_CREATE_TRANSACTIONS, has_operator_level_access)
+rules.add_perm(PERMISSION_CAN_READ_SUBSIDIES, has_admin_level_access)
+rules.add_perm(PERMISSION_CAN_READ_TRANSACTIONS, has_learner_level_access)
```
iloveagent57, line 140: very nice

enterprise_subsidy/apps/api/v1/views.py
```diff
@@ -1 +1,216 @@
...
+        permission_for_action = {
+            "list": PERMISSION_CAN_READ_SUBSIDIES,
+            "retrieve": PERMISSION_CAN_READ_SUBSIDIES,
+            "create": PERMISSION_NOT_GRANTED,
+            "update": PERMISSION_NOT_GRANTED,
+            "partial_update": PERMISSION_NOT_GRANTED,
+            "destroy": PERMISSION_NOT_GRANTED,
+        }
+        permission_required = permission_for_action.get(self.request_action, PERMISSION_NOT_GRANTED)
+        return permission_required
```
iloveagent57, line 56: This is super clean and helpful; there's probably a pattern we can extract from this and add to `edx_rbac.mixins`

enterprise_subsidy/apps/api/v1/views.py
```diff
@@ -1 +1,216 @@
...
+        response = {'message': 'Update action is not offered.'}
+        return Response(response, status=status.HTTP_403_FORBIDDEN)
+
+    def destroy(self, *args, **kwargs):
+        """
+        We should never get to this point because the permission for this action is not granted, but just in case we'll
+        force a 403 response.
+        """
+        response = {'message': 'Destroy action is not offered.'}
+        return Response(response, status=status.HTTP_403_FORBIDDEN)
```
iloveagent57, line 126: I like the belt+suspenders approach, here.  Another thing worth considering as an addition to edx-rbac.

iloveagent57: On the other hand, DRF does provide: https://www.django-rest-framework.org/api-guide/viewsets/#readonlymodelviewset which I think is actually what we want, here.

enterprise_subsidy/apps/api/v1/views.py
```diff
@@ -1 +1,216 @@
...
```
iloveagent57, line 216: You can probably use this example mixin inheritance to achieve the same thing: https://www.django-rest-framework.org/api-guide/viewsets/#custom-viewset-base-classes

iloveagent57 (commented):
This looks really great, @pwnage101 
We should chat a little bit more about the multi-table vs. single-table inheritance decision.  We should probably identify the most common query patterns we'll need to support, that should make the decision a little more obvious.

enterprise_subsidy/apps/api/v1/urls.py
```diff
@@ -1,4 +1,30 @@
-""" API v1 URLs. """
+"""
+URL definitions for enterprise-subsidy API version 1.
+
+Listing, reading, and testing subsidies::
+
+  GET /api/v1/subsidies/?enterprise_customer_uuid={uuid}&subsidy_type={"learner_credit","subscription"}
```
johnnagro, line 6: do we want to make `enterprise_customer_uuid` a path param rather than a query param? seems more restful to me, ie `/api/v1/subsidies/<enterprise_customer_uuid>/<subsidy_uuid>?subsidy_type=foo`

enterprise_subsidy/apps/api/v1/views.py
```diff
@@ -1 +1,180 @@
...
+    viewsets.GenericViewSet,
+):
+    """
+    ViewSet for the Transaction model.
+
+    Currently, this only supports listing, retrieving, creating, and reversing transactions::
+
+      GET  /api/v1/transactions/
+      GET  /api/v1/transactions/{transaction_uuid}/
+      POST /api/v1/transactions/?subsidy_uuid={subsidy_uuid}
```
johnnagro, line 116: to be more restful, `subsidy_uuid` would probably be a body/post param rather than a query param

enterprise_subsidy/apps/api/v1/views.py
```diff
@@ -1 +1,180 @@
...
+
+class SubsidyViewSet(
+    PermissionRequiredForListingMixin, mixins.RetrieveModelMixin, mixins.ListModelMixin, viewsets.GenericViewSet
+):
+    """
+    ViewSet for the Subsidy model.
+
+    Currently, this only supports listing, reading, and testing subsidies::
+
+      GET /api/v1/subsidies/?enterprise_customer_uuid={uuid}&subsidy_type={"learner_credit","subscription"}
```
johnnagro, line 32: similarly, should `enterprise_customer_uuid` be a path param here?

johnnagro (approved):
looks good, have some questions about how we mix path, query, and body params to make it more restful

enterprise_subsidy/apps/api/v1/views.py
iloveagent57: For reads: it feels more natural, and perhaps necessary, to list transactions for a particular subsidy.
We might want to split out a viewset for reading transactions into something like `TransactionReadViewset`, which can use `PermissionRequiredForListingMixin`, and have the URL path be scoped to the subsidy, i.e. `GET /api/v1/subsidies/[subsidy_uuid/transactions`, which feels more natural to me.  

Then something like `TransactionWriteViewset` to support creation and reversals.  This might make it easier to support more natural paths for creation and reversal, i.e. `POST /api/v1/subsidies/{subsidy_uuid}/transactions`  to create a transaction (although maybe we don't necessarily need the subsidy uuid as part of the path for reversing a transaction, since you'd have the `transaction_uuid`, which is the PK of the thing to reverse).

Maybe some of these endpoints can be plain old views that use the `@permission_required` rbac decorator?

iloveagent57 (approved):
We can definitely move forward with this and come back to some of the REST API design questions as a follow-up.

PR #15, opened 2023-02-27
feat: connect POST /api/v1/transactions/ to redeem()

pwnage101:
Before this change, POST /api/v1/transactions/ inherits the default create action to create a Transaction.  After this change, it is actually hooked up to the redeem() function which contains the rest of the business logic to redeem a subsidy, aside from just the creation of a Transaction.

ENT-6836


pwnage101:
Tests pass locally, I just have to bump openedx_ledger for github to pass.

johnnagro (approved):
⛵

PR #36, opened 2023-03-20
feat: add subsidy can_redeem action to API

iloveagent57:
### Description
* Adds a `can_redeem` detail action to the `SubsidyViewSet`
* More strongly formalizes what it means to be redeemable.
* Adds unit tests for all the functions that deal with redeemability.

### Testing instructions

Add some, if applicable

### Merge checklist
- [x] All reviewers approved
- [x] CI build is green
- [x] Documentation updated (not only docstrings)
- [x] Commits are squashed

### DIFF
--- a/enterprise_subsidy/apps/api/v1/views/subsidy.py
+++ b/enterprise_subsidy/apps/api/v1/views/subsidy.py
@@ -2,13 +2,17 @@
 Views for the enterprise-subsidy service relating to the Subsidy model
  service.
 """
+from django.utils.functional import cached_property
 from edx_rbac.mixins import PermissionRequiredForListingMixin
 from edx_rest_framework_extensions.auth.jwt.authentication import JwtAuthentication
-from rest_framework import mixins, permissions, viewsets
+from rest_framework import mixins, permissions, status, viewsets
 from rest_framework.authentication import SessionAuthentication
+from rest_framework.decorators import action
+from rest_framework.response import Response
 
 from enterprise_subsidy.apps.api.v1 import utils
-from enterprise_subsidy.apps.api.v1.serializers import SubsidySerializer
+from enterprise_subsidy.apps.api.v1.serializers import SubsidySerializer, TransactionSerializer
+from enterprise_subsidy.apps.subsidy.api import can_redeem
 from enterprise_subsidy.apps.subsidy.constants import (
     ENTERPRISE_SUBSIDY_ADMIN_ROLE,
     ENTERPRISE_SUBSIDY_OPERATOR_ROLE,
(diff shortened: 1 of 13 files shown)


enterprise_subsidy/apps/subsidy/api.py
```diff
@@ -38,3 +40,43 @@ def get_or_create_learner_credit_subsidy(
         defaults=subsidy_defaults,
     )
     return (subsidy, created)
+
+
+def can_redeem(subsidy, learner_id, content_key):
```
adamstankiewicz, line 45: tiny naming nit: perhaps `lms_user_id` over `learner_id` (to avoid any confusion of whether this means an `EnterpriseCustomerUser` id), though it is described in the docstring below. Similar feedback throughout.

enterprise_subsidy/apps/subsidy/api.py
```diff
@@ -38,3 +40,43 @@ def get_or_create_learner_credit_subsidy(
...
+    if existing_transaction and not _existing_reversal(existing_transaction):
+        is_redeemable = False
+        price_for_content = subsidy.price_for_content(content_key)
+    else:
+        is_redeemable, price_for_content = subsidy.is_redeemable(content_key)
+
+    return (is_redeemable, price_for_content, existing_transaction)
+
+
+def _existing_reversal(transaction):
```
adamstankiewicz, line 74: [curious/clarification] Very edge case-y, but can a transaction's reversal ever be reversed itself (i.e., "undo" a refund), where it _would_ count towards a prior redemption? If so, does this method take this into account? Probably something we can defer on.

iloveagent57: I don't think the models support that today.  To undo a refund, we'd probably have to just make a new redemption/create a new transaction.

adamstankiewicz: 👍 Cool. That's probably a good thing the models don't support that today; less complexity to deal with 😄

alex-sheehan-edx (approved):
lgtm

adamstankiewicz (approved):
LGTM with clarifying question around why a `POST` vs. a `GET`.

enterprise_subsidy/apps/api/v1/views/subsidy.py
```diff
@@ -104,3 +119,54 @@ def base_queryset(self):
             "ledger__transactions",
             "ledger__transactions__reversal",
         ).order_by("uuid")
+
+    @action(methods=['post'], detail=True)
```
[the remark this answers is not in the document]
pwnage101: this is defined as a GET in the initial API spec, FWIW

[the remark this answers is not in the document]
iloveagent57: ah, yup, good call

enterprise_subsidy/apps/api/v1/views/subsidy.py
```diff
@@ -104,3 +119,54 @@ def base_queryset(self):
...
+
+        redeemable, content_price, existing_transaction = can_redeem(
+            self.requested_subsidy,
+            lms_user_id,
+            content_key,
+        )
+        response_payload = {
+            'can_redeem': redeemable,
+            'content_price': content_price,
+            'unit': UnitChoices.USD_CENTS,
```
pwnage101, line 166: Why not pass the unit of the subsidy?

pwnage101 (approved):
LGTM!  docs nits.

PR #39, opened 2023-03-22
feat: drf-spectacular stuff

iloveagent57:
### Description
Still a WIP, but hopefully this gives an idea of the usefulness of drf-spec.

check out this code, then try the redoc page: http://localhost:18280/api/schema/redoc/#tag/subsidy/operation/api_v1_subsidies_can_redeem_retrieve

Notice how OAS tags have helped organize the endpoints:
image

Then `extend_schema()` helps us do a ton of stuff, like
formalize the query params:
image

formalize the response schema through the use of a custom response serializer:
image

and then that response serializer also helps provide a good example of the serialized response data, in this case, showing the nested serialization of a Transaction record:
image

### Testing instructions

Add some, if applicable

### Merge checklist
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Commits are squashed


johnnagro:
I like your ideas and wish to subscribe to your newsletter.

adamstankiewicz:
This is awesome! Looking at the docs for drf-spectacular, it seems like it works with both SwaggerUI and ReDoc. Is the idea that this can integrate with our existing Swagger docs at `/api-docs` or is your recommendation we'd start using ReDoc instead of Swagger moving forward?

iloveagent57:
@adamstankiewicz There's a v2 of Swagger docs at http://localhost:18280/api/schema/swagger-ui/

I want to deprecate our current swagger UI eventually in favor of this better swagger UI, just need to figure out if there's any downstream dependencies, update docs, communicate, etc.

iloveagent57:
I think as long as there's no (or very minimal) extra work to support both, vs. supporting only redoc, we should support both so that devs can pick the tool that feels best for them. I share your preference for ReDoc.

Swagger does support the real request option, but I think we'd have to put some amount of work into making sure the authentication and permissions work as intended.

macdiesel (approved):

PR #44, opened 2023-03-30
feat: add revenue_category field to Subsidy.

pwnage101:
This new field is a spiritual replacement of the slug of a Catalogue.Category assigned to a Voucher.CouponVouchers object. To start, it's primary purpose is to control downstream revenue recognition logic, and otherwise has no impact to the application logic.

I chose to call it a "category" rather than a "type" in order to emphasize that a subsidies are bucketed by this value.

Also:
* Be more opinionated about what internal_only does. This hopefully forces us to either accept it or reject it as a feature.
* Add help texts to all other Subsidy fields.
* More clearly prepare Subsidy for future split into LearnerCreditSubsidy and SubscriptionSubsidy by reserving the "subsidy_type" field name.

ENT-6841


enterprise_subsidy/apps/subsidy/models.py
```diff
@@ -76,15 +107,19 @@ class Subsidy(TimeStampedModel):
...
+        null=False,
+        blank=False,
+        help_text="The positive balance this Subidy will be initially provisioned to start with."
     )
     ledger = models.OneToOneField(
         Ledger,
         null=True,
         blank=True,
         on_delete=models.SET_NULL,
+        help_text="The one and only one Ledger uniquely associated with this Subsidy."
```
iloveagent57, line 122: nit: one-too-many "ones"

iloveagent57 (approved):

PR #45, opened 2023-04-03
fix: content-metadata endpoint now returns price in cents as a float.

iloveagent57:
### Description
Always return price as a float, in cents.


johnnagro (approved):
⛵

PR #48, opened 2023-04-03
feat: update enum for subsidy reference type to inclue a SF OpportunityLineItem

pwnage101:
We have been informed that this is the correct object in salesforce to use as a subsidy reference.


enterprise_subsidy/apps/subsidy/models.py
```diff
@@ -44,9 +44,9 @@ class SubsidyReferenceChoices:
     object that caused the subsidy to come into existence.
     """
     # TODO: confirm the actual Salesforce object to use, I sort of just made up "opportunity_product_id"...
-    OPPORTUNITY_PRODUCT_ID = "opportunity_product_id"
+    OPPORTUNITY_LINE_ITEM = "salesforce_OpportunityLineItem"
```
iloveagent57, line 47: nit: we should snake-case this value to make it consistent with how we typically define the persisted value of choices.

iloveagent57 (changes requested):

pwnage101:
* removed outdated comment
* prefixed enum key with `SALESFORCE_` to be more explicit
* made db value fully snake_case

iloveagent57 (approved):

PR #51, opened 2023-04-05
refactor: move content metadata stuff to content_metadata/api.py

iloveagent57:
### Description
A little bit of pre-factoring of content metadata functions that'll be used in the fulfillment module.

https://2u-internal.atlassian.net/browse/ENT-6804

### Testing instructions

No change in functionality, just run the unit tests.

### Merge checklist
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Commits are squashed


enterprise_subsidy/apps/content_metadata/api.py
```diff
@@ -2,6 +2,75 @@
...
+
+def price_for_content(content_data):
+    """
+    Helper to return the "official" price for content.
+    The endpoint at ``self.content_metadata_url`` will always return price fields
+    as USD (dollars), possibly as a string or a float.  This method converts
+    those values to USD cents as an integer.
+    """
+    content_price = None
+    if content_data.get('first_enrollable_paid_seat_price'):
```
johnnagro, line 29: this isn't covered in the current tests - is this really an expected form of price? if so we may wanna add a test

enterprise_subsidy/apps/content_metadata/api.py
```diff
@@ -2,6 +2,75 @@
...
+
+    if not content_price:
+        enrollment_mode_for_content = mode_for_content(content_data)
+        for entitlement in content_data.get('entitlements', []):
+            if entitlement.get('mode') == enrollment_mode_for_content:
+                content_price = entitlement.get('price')
+
+    if content_price:
+        return int(Decimal(content_price) * CENTS_PER_DOLLAR)
+    return None
```
johnnagro, line 40: `None` is fine here but i also am wondering if we'd want an exception to be raised - would that be overkill?

johnnagro (approved):
⛵

enterprise_subsidy/apps/fulfillment/api.py
```diff
@@ -1,34 +1,117 @@
...
+
+
+def get_fulfillment_handlers(content_metadata):
+    """
+    Could implement this as multiple fulfillment operations, but all on a single
+    transaction record to fulfill.
     """
-    raise NotImplementedError
+    if content_metadata.get('product_source') == 'the-twou-product-source':
+        return [GEAGFulfillmentHandler(content_metadata), OpenCourseFulfillmentHandler(content_metadata)]
```
johnnagro, line 117: the `OpenCourseFulfillmentHandler` is different than other fulfillment providers because our "commit transaction" relies on the enterprise enrollment. i think we should just call this explicitly every time and not include it in a manager like this

johnnagro:
ive broken out most of this work into dedicated PRs

PR #54, opened 2023-04-13
feat: Take ledger locking into account

pwnage101:
This is a followup from openedx_ledger changes in https://github.com/openedx/openedx-ledger/pull/17 to implement locking on the ledger object during transaction creation.

* POST /api/v1/transactions: add a new 429 response to  when the ledger is locked.
* POST /api/v1/transactions: update reponse code to 422 (was 403) when subsidy is not redeemable.
* Add basic model test for Subsidy.redeem(), and additional view tests to cover the above two bullets.
* Also, make sure the Transaction lifecycle fully progressess through either:
   - created -> pending -> committed
   - created -> pending -> failed
* make upgrade
  - remove codecov package from ci.txt

ENT-6835

### DIFF
--- a/enterprise_subsidy/apps/api/v1/tests/test_views.py
+++ b/enterprise_subsidy/apps/api/v1/tests/test_views.py
@@ -9,12 +9,13 @@ from functools import partial
 from unittest import mock
 
 import ddt
-from openedx_ledger.models import TransactionStateChoices
+from openedx_ledger.models import Transaction, TransactionStateChoices
 from openedx_ledger.test_utils.factories import (
     ExternalFulfillmentProviderFactory,
     ExternalTransactionReferenceFactory,
     TransactionFactory
 )
+from requests.exceptions import HTTPError
 from rest_framework import status
 from rest_framework.reverse import reverse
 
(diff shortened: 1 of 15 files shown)


enterprise_subsidy/apps/api/v1/tests/test_views.py
```diff
@@ -687,15 +688,15 @@ def test_retrieve_invalid_uuid(self):
...
+    @mock.patch("enterprise_subsidy.apps.subsidy.models.Subsidy.price_for_content")
+    def test_create(self, mock_price_for_content, mock_enterprise_client):
         """
         Test create Transaction, happy case.
         """
         url = reverse("api:v1:transaction-list")
         test_enroll_enterprise_fulfillment_uuid = "test-enroll-reference-id"
         mock_enterprise_client.enroll.return_value = test_enroll_enterprise_fulfillment_uuid
-        mock_get_course_price.return_value = "100.00"
+        mock_price_for_content.return_value = 10000
```
pwnage101, line 699: This was a pretty sneaky bug that caused tests to be non-deterministic.  Basically, if a function is cached (e.g. via the `@lru_cache()` decorator), it will cache between unit tests unless we mock the function itself which overrides the cache.

iloveagent57: 👍 good catch.

iloveagent57 (approved):
Looks good, minor quibble about HTTP response code, but not the ant-hill I'm gonna die on.

enterprise_subsidy/apps/api/v1/views/transaction.py
```diff
@@ -386,11 +389,17 @@ def create(self, request, *args, **kwargs):
...
             rest_framework.response.Response:
-                400: If there are missing or otherwise invalid input parameters.  Response body is JSON with a single
+                400: There are missing or otherwise invalid input parameters.  Response body is JSON with a single
+                     `Error` key.
+                403: The requester has insufficient create permissions.  Response body is JSON with a single `Error`
+                     key.
+                422: The subisdy is not redeemable in a way that IS NOT retryable (e.g. the balance is too low, or
+                     content is not in catalog, etc.).  Response body is JSON with a
+                     single `Error` key.
+                429: The subisdy is not redeemable in a way that IS retryable (e.g. something else is already holding a
```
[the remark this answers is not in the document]
pwnage101: My reservation about using 422 was mainly that we could distinguish between retryable vs. non-retryable by treating 422 as non-retryable and using something else for retryable.

My reservation about using 429 was that same statement from mozilla claiming that browsers may just interpret 429 as equivalent to 400 because it should never happen in the first place.

PR #55, opened 2023-04-13
fix: content_key not found should respond with 422, and stop double-converting dollars to cents

pwnage101:
This combo commit combines two unrelated fixes that are too hard to split up into two commits because they both touch the same lines of code:

* On redeem (create transaction) if the requested content_key is not found, the response should be 422 instead of 500.
* Stop double-converting dollars to cents.  Previously, transaction records would be written with quantity values 100x larger than they should have been.

Unticketed work.

### DIFF
--- a/enterprise_subsidy/apps/api/v1/tests/test_views.py
+++ b/enterprise_subsidy/apps/api/v1/tests/test_views.py
@@ -141,10 +141,6 @@ class SubsidyViewSetTests(APITestBase):
     get_list_url = partial(reverse, "api:v1:subsidy-list")
     get_can_redeem_url = partial(reverse, "api:v1:subsidy-can-redeem")
 
-    def setUp(self):
-        super().setUp()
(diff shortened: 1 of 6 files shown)


enterprise_subsidy/apps/api/v1/tests/test_views.py
```diff
@@ -141,10 +141,6 @@ class SubsidyViewSetTests(APITestBase):
     get_list_url = partial(reverse, "api:v1:subsidy-list")
     get_can_redeem_url = partial(reverse, "api:v1:subsidy-can-redeem")
 
-    def setUp(self):
-        super().setUp()
-        self.subsidy_1.catalog_client = mock.MagicMock()
-
```
pwnage101, line 147: This wasn't doing anything anyway because the subsidy model instance from the unit test's perspective is a different instance than from the view's perspective, despite them both referring to the same DB object.

iloveagent57 (approved):

PR #57, opened 2023-04-19
Aed/refactor tx view

iloveagent57:
### Description
* Creates a v2 API path for listing user-scoped transactions.
* Createa a v2 API path in which admins can list transactions for a subsidy
* Creates (the same) v2 path with which operators can create transactions.
* installs django-filter for handling filtering query params in a more standardized way
* installs helpful django-extensions and debug toolbar tools.
* demos how to set filtering and filterset classes correctly, so that drf-spec will pick up the model field help_text.

https://2u-internal.atlassian.net/browse/ENT-6947

### Testing instructions

Add some, if applicable

### Merge checklist
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Commits are squashed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -164,13 +164,16 @@ docker_build:
 # devstack-themed shortcuts
 dev.up: dev.up.redis
 	docker-compose up -d
+	docker exec -u 0 -it enterprise-subsidy.app pip-sync -q requirements/dev.txt requirements/private.* requirements/test.txt
 
 dev.up.build: dev.up.redis
 	docker-compose up -d --build
+	docker exec -u 0 -it enterprise-subsidy.app pip-sync -q requirements/dev.txt requirements/private.* requirements/test.txt
 
 dev.up.build-no-cache: dev.up.redis
 	docker-compose build --no-cache
 	docker-compose up -d
+	docker exec -u 0 -it enterprise-subsidy.app pip-sync -q requirements/dev.txt requirements/private.* requirements/test.txt
 
 dev.up.redis: # This has the nice side effect of starting the devstack_default network
 	docker-compose -f $(DEVSTACK_WORKSPACE)/devstack/docker-compose.yml up -d redis
--- a/enterprise_subsidy/apps/api/v1/serializers.py
+++ b/enterprise_subsidy/apps/api/v1/serializers.py
@@ -1,11 +1,14 @@
 """
 Serializers for the enterprise-subsidy API.
 """
+from logging import getLogger
 
-from openedx_ledger.models import Reversal, Transaction
+from openedx_ledger.models import LedgerLockAttemptFailed, Reversal, Transaction
 from rest_framework import serializers
 
 from enterprise_subsidy.apps.subsidy.models import Subsidy
+
+logger = getLogger(__name__)
 
 
 class SubsidySerializer(serializers.ModelSerializer):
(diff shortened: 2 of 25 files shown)


enterprise_subsidy/apps/api/schema.py
```diff
@@ -8,12 +8,61 @@
...
+            'location': OpenApiParameter.QUERY,
+            'required': True,
+            'allow_blank': False,
+            'description': "The uuid of the subsidy access policy to which the query pertains.",
+        }
+        options.update(kwargs)
+        return super().__init__(QueryParams.subsidy_access_policy_uuid, **options)
+
+
 class Parameters:
```
pwnage101, line 60: Parameters class is dead code now, right?

enterprise_subsidy/apps/api/v1/serializers.py
```diff
@@ -106,6 +109,67 @@ def get_unit(self, obj):
         return obj.ledger.unit if obj.ledger else None
 
 
+class TransactionCreationError(Exception):
+    """
+    Generic exception related to errors during transaction creation.
+    """
+
+
+class CreateTransactionSerializer(serializers.ModelSerializer):
```
pwnage101, line 118: My favorite takeaway from the enterprise-access code was finding either "Request" or "Response" in the name of the serializer, which really helped me understand what the serializer was actually used for.  Is that a general pattern you like, and should we adopt it here by calling it CreateTransactionRequestSerializer?

enterprise_subsidy/apps/api/v1/serializers.py
```diff
@@ -106,6 +109,67 @@ def get_unit(self, obj):
...
+            'lms_user_id',
+            'content_key',
+            'subsidy_access_policy_uuid',
+        ]
+        # Make lms_user_id, content_key, and subsidy_access_policy_uuid all required
+        extra_kwargs = {
+            'idempotency_key': {'required': False},
+            'lms_user_id': {'required': True},
+            'content_key': {'required': True},
+            'subsidy_access_policy_uuid': {'required': True},
```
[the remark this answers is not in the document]
iloveagent57: Yeah, good point, lemme take another crack at that - it should be possible to define it only on the serializer(s).

enterprise_subsidy/apps/api/schema.py
iloveagent57: `Parameters` still is used by the v1 `can_redeem` endpoint.  I removed the code above, though, once I found a way to utilize `FilterSet` in a way that drf-spec can usefully introspect.

enterprise_subsidy/apps/api/v1/serializers.py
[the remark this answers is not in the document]
iloveagent57: ok, so on further reflection, here's where I landed, with a primary goal being that the `help_text` defined on the model fields auto-magically routes it's way into the API schema/docs generated by DRF spec, in a way that's relatively DRY.
1. I got rid of those `OpenApiParameter()` definitions, and they're no longer included in the view.
2. If we want that nice "help_text floats up to API spec promised-land" property, we actually need to override the `required` value for this creation serializer using `extra_kwargs` in the Meta class - declaring the field explicitly with something like
```python
idempotency_key = serializers.CharField(required=False, allow_null=True)
```
loses all the other metadata about the field that originates in the model definition.
3. So in the current iteration of this PR, you'll only have to look in the serializer for the create action to understand or declare a field as required, it's just in a not-completely-obvious place.  But I'm OK with this.

Furthermore, to get all the nice field metadata properties wired up to the API schema for the list action, I added an explicit `FilterSet` class that drf-spec can usefully introspect.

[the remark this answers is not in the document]
pwnage101: That all sounds good to me!

pwnage101 (approved):

enterprise_subsidy/apps/api/v2/views/transaction.py
```diff
@@ -0,0 +1,202 @@
...
+
+# https://drf-spectacular.readthedocs.io/en/latest/faq.html#using-extend-schema-on-apiview-has-no-effect
+# @extend_schema needs to be applied to the entrypoint method of the view.
+# For views derived from Viewset, these are methods like retrieve, list, create.
+# For APIView based views, these are get, post, create.
+# ListAPIView does in fact have a list method (via mixin), but the actual entrypoint
+# is still the get method, and the list call is proxied through the entrypoint.
+# Similarly for post/create.
+@extend_schema_view(
+    get=extend_schema(
```
pwnage101, line 79: Nit: I feel like the code read faster if we put extend_schema on the view methods directly, otherwise you have to do a little bit of mental gymnastics to connect get -> list and create -> post

enterprise_subsidy/apps/api/v2/views/transaction.py
```diff
@@ -0,0 +1,202 @@
...
+        """
+        Returns the requested ``subsidy_uuid`` path parameter.
+        """
+        return self.kwargs.get('subsidy_uuid')
+
+    @cached_property
+    def subsidy(self):
+        """
+        Returns the Subsidy instance from the requested ``subsidy_uuid``.
+        """
```
pwnage101, line 55: Nit: add documentation that this raises Subsidy.DoesNotExist (which bubbles up as a 404 response?)

pwnage101 (approved):
more nits!

enterprise_subsidy/apps/api/v2/views/transaction.py
iloveagent57: the auth is actually going to fail first and result in a 403, I think all the code is now updated to reflect that.

enterprise_subsidy/apps/api/v2/tests/test_transaction_views.py
```diff
@@ -0,0 +1,560 @@
...
+        """
+        self.set_up_admin()
+        self.set_jwt_cookie([
+            (SYSTEM_ENTERPRISE_ADMIN_ROLE, ALL_ACCESS_CONTEXT),
+            (SYSTEM_ENTERPRISE_LEARNER_ROLE, self.enterprise_1_uuid),
+        ])
+
+        # The all-access admin role assignment should let the admin
+        # user read the transactions for subsidy_3
+        url = reverse("api:v2:transaction-admin-list-create", args=[self.subsidy_3_uuid])
```
pwnage101, line 328: I think the correct test here would be whatever makes the subsidy under test have the same enterprise customer as what's paired with the learner role in the JWT.  So, either change the learner context to `self.enterprise_2_uuid`, or change the subsidy under test to `self.subsidy_1_uuid`.

enterprise_subsidy/apps/api/v2/tests/test_transaction_views.py
```diff
@@ -0,0 +1,560 @@
...
+        Neither learner or admin roles should provide the ability to create transactions.
+        """
+        if role == 'admin':
+            self.set_up_admin()
+        if role == 'learner':
+            self.set_up_learner()
+
+        url = reverse("api:v2:transaction-admin-list-create", args=[self.subsidy_1.uuid])
+
+        response = self.client.post(url, {'anything': 'goes'})
```
pwnage101, line 402: IMO the post parameters for a test like this should be valid to narrow the scope of what's being tested

enterprise_subsidy/apps/api/v2/tests/test_transaction_views.py
```diff
@@ -0,0 +1,560 @@
...
+    def test_operator_creation_required_fields_validation_eror(self):
+        """
+        Tests that an authenticated operator receives a 400 response
+        when attempting to create without all of the required fields being present.
+        """
+        self.set_up_operator()
+
+        url = reverse("api:v2:transaction-admin-list-create", args=[self.subsidy_1.uuid])
+
+        with mock.patch.object(self.subsidy_1, 'redeem', side_effect=LedgerLockAttemptFailed):
```
pwnage101, line 458: locking code copy-pasta from the last test?

Makefile
```diff
@@ -164,13 +164,16 @@ docker_build:
 # devstack-themed shortcuts
 dev.up: dev.up.redis
 	docker-compose up -d
+	docker exec -u 0 -it enterprise-subsidy.app pip-sync -q requirements/dev.txt requirements/private.* requirements/test.txt
```
pwnage101, line 167: good idea!

pwnage101 (commented):
The rest LGTM!  I like the overall approach, the auth is way more robust.

Makefile
johnnagro: this could be a make command, which is chained on and thus re-used rather than redefined?

johnnagro (approved):
looks one, one nit.

PR #60, opened 2023-04-20
feat: add geag variant to contentmetadata

johnnagro:
### Description

- Breaking out some of the work in https://github.com/openedx/enterprise-subsidy/pull/51
- refactor some of the content metadata stuff
- add geag variant id capture

### Testing instructions

Add some, if applicable

### Merge checklist
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Commits are squashed

### DIFF
--- a/enterprise_subsidy/apps/api/v1/views/content_metadata.py
+++ b/enterprise_subsidy/apps/api/v1/views/content_metadata.py
@@ -18,6 +18,7 @@ from rest_framework.status import HTTP_404_NOT_FOUND
 from enterprise_subsidy.apps.api.v1 import utils
 from enterprise_subsidy.apps.api.v1.decorators import require_at_least_one_query_parameter
 from enterprise_subsidy.apps.api_client.enterprise_catalog import EnterpriseCatalogApiClient
+from enterprise_subsidy.apps.content_metadata.api import summary_data_for_content
 from enterprise_subsidy.apps.subsidy.constants import (
     ENTERPRISE_SUBSIDY_ADMIN_ROLE,
     ENTERPRISE_SUBSIDY_LEARNER_ROLE,
@@ -99,7 +100,7 @@ class ContentMetadataViewSet(
                 enterprise_customer_uuid[0],
                 content_identifier
             )
-            content_summary = catalog_client.summary_data_for_content(content_data)
+            content_summary = summary_data_for_content(content_data)
             if not content_summary.get('content_price'):
                 logger.warning("Could not find course price in metadata for {content_identifier}")
         except requests.exceptions.HTTPError as exc:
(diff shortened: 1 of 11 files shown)


enterprise_subsidy/apps/api/v1/views/content_metadata.py
```diff
@@ -99,7 +100,7 @@ def get(self, request, content_identifier, enterprise_customer_uuid):
                 enterprise_customer_uuid[0],
                 content_identifier
             )
-            content_summary = catalog_client.summary_data_for_content(content_data)
+            content_summary = summary_data_for_content(content_data)
```
pwnage101, line 103: Why not keep this a catalog_client method?  All other usages of content_metadata helper functions are hidden behind catalog_client, this seems to deviate from that pattern.

enterprise_subsidy/apps/content_metadata/constants.py
```diff
@@ -0,0 +1,7 @@
+"""
+Constants about content metadata.
+"""
+EDX_PRODUCT_SOURCE = "edX"
+TWOU_PRODUCT_SOURCE = "2u"
+EXECUTIVE_EDUCATION_MODE = "paid-executive-education"
+EDX_VERIFIED_COURSE_MODE = "verified"
```
pwnage101, line 7: Nit: should enum options be [grouped into classes](https://docs.python.org/3/howto/enum.html#using-a-descriptive-string)? something like:

```suggestion
from enum import Enum

class ProductSource(Enum)
    """
    TODO
    """
    EDX = "edX"  # e.g. OCM courses
    TWOU = "2u"  # e.g. ExecEd courses

etc.
```

enterprise_subsidy/apps/content_metadata/api.py
```diff
@@ -2,6 +2,86 @@
...
+    Helps get the product source string, given a dict of ``content_data``.
+    """
+    if product_source := content_data.get('product_source'):
+        source_name = product_source.get('name')
+        if source_name in CONTENT_MODES_BY_PRODUCT_SOURCE:
+            return source_name
+    return EDX_PRODUCT_SOURCE
+
+
+def get_geag_varient_id_for_content(content_data):
```
pwnage101, line 62: varient -> variant

enterprise_subsidy/apps/content_metadata/api.py
```diff
@@ -2,6 +2,86 @@
...
+    """
+    Returns the GEAG ``variant_id`` or ``None``, given a dict of ``content_data``.
```
pwnage101, line 64: Can you describe what the GEAG variant_id is in this docstring?

enterprise_subsidy/apps/api/v1/views/content_metadata.py
johnnagro: i believe this clarifies the separation of concerns - allowing the catalog client to mostly just be how we interact with the catalog service, and the new metadata api focus on how subsidy service is going to process/transform the catalog data into what it needs - how does that sound?

pwnage101: That idea sounds good to me, but it feels like this PR only partially commits to that idea because some processing of the metadata is still happening inside the catalog_client (extracting the price).  It would feel more like concerns have been separated if catalog_client exposed more of a 1:1 of class methods to API endpoints.

pwnage101 (approved):

PR #66, opened 2023-04-28
refactor: extract transaction list aggregates into a shareable paginator class

iloveagent57:
### Description
Will make it easy to plug the aggregates pagination into other versions of transaction views.

### Testing instructions

Add some, if applicable

### Merge checklist
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Commits are squashed

### DIFF
--- a/enterprise_subsidy/apps/api/v1/views/transaction.py
+++ b/enterprise_subsidy/apps/api/v1/views/transaction.py
@@ -17,6 +17,7 @@ from rest_framework.authentication import SessionAuthentication
 from rest_framework.exceptions import ParseError
 from rest_framework.response import Response
 
+from enterprise_subsidy.apps.api.paginators import TransactionListPaginator
 from enterprise_subsidy.apps.api.v1 import utils
 from enterprise_subsidy.apps.api.v1.decorators import require_at_least_one_query_parameter
 from enterprise_subsidy.apps.api.v1.serializers import TransactionSerializer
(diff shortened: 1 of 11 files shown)


enterprise_subsidy/apps/api/v1/views/transaction.py
```diff
@@ -361,21 +362,9 @@ def list(self, request, *args, **kwargs):
...
-                "remaining_subsidy_balance": subsidy.current_balance(),
-                "total_quantity": subsidy.ledger.subset_balance(self.get_queryset()),
-            }
-            # Compose a response that combines the base response with additional aggregates data.
-            response_data = {"aggregates": aggregates}
-            response_data.update(base_response.data)
-            return Response(response_data, status=base_response.status_code, headers=base_response.headers)
-        else:
-            return base_response
+            self._paginator = TransactionListAggregatesPaginator()
```
iloveagent57, line 366: I realize that's a little magical, but DRF doesn't provide a good way to dynamically set the pagination class within a viewset.
https://github.com/encode/django-rest-framework/issues/6397

pwnage101 (approved):

enterprise_subsidy/apps/api/v1/views/transaction.py
```diff
@@ -361,21 +362,9 @@ def list(self, request, *args, **kwargs):
...
-                "remaining_subsidy_balance": subsidy.current_balance(),
-                "total_quantity": subsidy.ledger.subset_balance(self.get_queryset()),
-            }
-            # Compose a response that combines the base response with additional aggregates data.
-            response_data = {"aggregates": aggregates}
-            response_data.update(base_response.data)
-            return Response(response_data, status=base_response.status_code, headers=base_response.headers)
-        else:
-            return base_response
+            self._paginator = TransactionListAggregatesPaginator()  # pylint: disable=attribute-defined-outside-init
```
johnnagro, line 366: i suppose one could also make a paginator class which does this routing, calling `TransactionListAggregatesPaginator` when appropriate and something else otherwise. That way one could just define the paginator normally without this call into the (protected) `_paginator`

johnnagro (approved):
looks okay - one consideration was added

PR #68, opened 2023-05-01
feat: geag fulfillment

johnnagro:
### Description

- Adds a fulfillment handler for GetSmarter Exec Ed fulfillment.
- https://2u-internal.atlassian.net/browse/ENT-6804
- https://2u-internal.atlassian.net/browse/ENT-6931
- https://2u-internal.atlassian.net/browse/ENT-7068
- https://2u-internal.atlassian.net/browse/ENT-7079

### Testing instructions

Add some, if applicable

### Merge checklist
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Commits are squashed

### DIFF
--- a/enterprise_subsidy/apps/content_metadata/api.py
+++ b/enterprise_subsidy/apps/content_metadata/api.py
@@ -150,3 +150,9 @@ class ContentMetadataApi:
             content_identifier
         )
         return self.product_source_for_content(course_details)
+
+    def get_geag_variant_id(self, enterprise_customer_uuid, content_identifier):
+        """
+        Returns the GetSmarter product variant id or None
+        """
+        return self.get_content_summary(enterprise_customer_uuid, content_identifier).get('geag_variant_id')
--- a/enterprise_subsidy/apps/fulfillment/api.py
+++ b/enterprise_subsidy/apps/fulfillment/api.py
@@ -7,6 +7,9 @@ from enterprise_subsidy.apps.content_metadata import api as content_metadata_api
 
 from .constants import EXEC_ED_2U_COURSE_TYPES, OPEN_COURSES_COURSE_TYPES
 
+from django.conf import settings
+from enterprise_subsidy.apps.content_metadata.api import ContentMetadataApi
+from getsmarter_api_clients.geag import GetSmarterEnterpriseApiClient
 
 def create_fulfillment(subsidy_uuid, lms_user_id, content_key, **metadata):
     """
@@ -32,3 +35,78 @@ def determine_fulfillment_client(subsidy_uuid, content_key):
     return None
     """
     raise NotImplementedError
+
+
+class GEAGFulfillmentHandler():
+    """
+    A class for fulfilling a GEAG transaction.
+    """
+    CENTS_PER_DOLLAR = 100.0
+
+    def get_smarter_client(self):
+        return GetSmarterEnterpriseApiClient(
+            client_id=settings.GET_SMARTER_OAUTH2_KEY,
+            client_secret=settings.GET_SMARTER_OAUTH2_SECRET,
+            provider_url=settings.GET_SMARTER_OAUTH2_PROVIDER_URL,
+            api_url=settings.GET_SMARTER_API_URL
+        )
+
+    def _get_geag_transaction_price(self, transaction):
+        return float(transaction.quantity) / self.CENTS_PER_DOLLAR
+
+    def _get_enterprise_customer_uuid(self, transaction):
+        return transaction.ledger.subsidy.enterprise_customer_uuid
+
+    def _get_geag_variant_id(self, transaction):
+        ent_uuid = self._get_enterprise_customer_uuid(transaction)
+        return ContentMetadataApi().get_geag_variant_id(ent_uuid, transaction.content_key)
+
+    def _create_allocation_payload(self, transaction, currency='USD'):
+        transaction_price = self._get_geag_transaction_price(transaction)
+        return {
+            'payment_reference': transaction.uuid,
+            'enterprise_customer_uuid': self.subsidy.enterprise_customer_uuid,
+            'currency': currency,
+            'order_items': [
+                {
+                    # productId will be the variant id from product details
+                    'productId': self._get_geag_variant_id(transaction),
+                    'quantity': 1,
+                    'normalPrice': transaction_price,
+                    'discount': 0.0,
+                    'finalPrice': transaction_price,
+                }
+            ],
+            'first_name': transaction.metadata.get('geag_first_name'),
+            'last_name': transaction.metadata.get('geag_last_name'),
+            'date_of_birth': transaction.metadata.get('geag_date_of_birth'),
+            'terms_accepted_at': transaction.metadata.get('geag_terms_accepted_at'),
+        }
+
+    def _validate(self, transaction):
+        """
+        Validates that a ledger transaction contains all the required information to
+        construct a GEAG allocation payload.
+
+        Raises an exception when the transaction is missing required information
+        """
+        for field in ['geag_first_name', 'geag_last_name', 'geag_date_of_birth', 'geag_terms_accepted_at']:
+            if not transaction.metadata.get(field):
+                # TODO make this a proper exception class
+                raise f'missing {field} transaction metadata'
+
+    def can_fulfill(self, transaction):
+        """
+        A helper to let callers know if the transaction at hand can be fulfilled
+        with this fulfillment handler.
+        """
+        return self._get_geag_variant_id(transaction)
+
+    def fulfill(self, transaction):
+        """
+        Attempt to fulfill a ledger transaction with the GEAG fulfillment system
+        """
+        self._validate(transaction)
+        allocation_payload = self._create_allocation_payload(transaction)
+        geag_response = self.get_smarter_client(**allocation_payload)
+        return geag_response.json().get('orderUuid')
--- a/enterprise_subsidy/apps/fulfillment/api.py
+++ b/enterprise_subsidy/apps/fulfillment/api.py
@@ -104,9 +104,10 @@ class GEAGFulfillmentHandler():
         for field in ['geag_first_name', 'geag_last_name', 'geag_date_of_birth', 'geag_terms_accepted_at']:
             if not transaction.metadata.get(field):
                 raise InvalidFulfillmentMetadataException(f'missing {field} transaction metadata')
+        return True
 
     def _save_fulfillment_reference(self, transaction, external_reference_id):
-        external_fulfillment_provider = ExternalFulfillmentProvider.objects.get_or_create(
+        external_fulfillment_provider, _ = ExternalFulfillmentProvider.objects.get_or_create(
             slug=self.EXTERNAL_FULFILLMENT_PROVIDER_SLUG
         )
         return ExternalTransactionReference.objects.create(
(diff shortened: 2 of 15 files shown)


enterprise_subsidy/apps/fulfillment/api.py
```diff
@@ -32,3 +36,92 @@ def determine_fulfillment_client(subsidy_uuid, content_key):
...
+
+    def _get_geag_variant_id(self, transaction):
+        ent_uuid = self._get_enterprise_customer_uuid(transaction)
+        return ContentMetadataApi().get_geag_variant_id(ent_uuid, transaction.content_key)
+
+    def _create_allocation_payload(self, transaction, currency='USD'):
+        transaction_price = self._get_geag_transaction_price(transaction)
+        return {
+            'payment_reference': transaction.uuid,
+            'enterprise_customer_uuid': self.subsidy.enterprise_customer_uuid,
```
iloveagent57, line 70: I think you want `transaction.ledger.subsidy` here, not `self.subsidy`, right?

enterprise_subsidy/apps/subsidy/models.py
```diff
@@ -380,9 +387,15 @@ def _create_redemption(
         ledger_transaction.state = TransactionStateChoices.PENDING
         ledger_transaction.save()
 
+        try:
+            if self.geag_fulfillment_handler().can_fulfill(ledger_transaction):
+                self.geag_fulfillment_handler().fulfill(ledger_transaction)
+        except Exception as exc:
+            self.rollback_transaction(ledger_transaction)
+            raise exc
```
iloveagent57, line 395: My only thought here is to make the exception handling more specific - we might not want to rollback the transaction depending on what point the fulfillment flow got to.  Doesn't have to block this PR.

enterprise_subsidy/apps/content_metadata/api.py
```diff
@@ -150,3 +150,9 @@ def get_product_source(self, enterprise_customer_uuid, content_identifier):
             content_identifier
         )
         return self.product_source_for_content(course_details)
+
```
iloveagent57, line 153: nit: Consider caching - doesn't have to block, but we should make sure there's a ticket in the backlog to eventually cache all this stuff at the right level of granularity.

iloveagent57 (commented):
Very reasonable

enterprise_subsidy/apps/subsidy/models.py
pwnage101: This is also the point at which we might take the path of returning a PENDING transaction all the way back to the enterprise-access -> frontend for async processing, right?

enterprise_subsidy/apps/subsidy/models.py
```diff
@@ -380,9 +387,15 @@ def _create_redemption(
...
+
         try:
             enterprise_fulfillment_uuid = self.enterprise_client.enroll(lms_user_id, content_key, ledger_transaction)
-            # TODO: once GEAG support is implemented we should pass external_reference (obj) to `commit_transactions`
```
pwnage101, line 385: comment removed, but not fulfilled

enterprise_subsidy/apps/subsidy/models.py
johnnagro: @iloveagent57 yep, agreed lemme noodle this further as the final iteration comes together.

@pwnage101 re: our IRL discussion earlier, we are proceeding with this being fairly synchronous for now.

enterprise_subsidy/apps/content_metadata/api.py
johnnagro: agreed. something i'll add into a later iteration.

enterprise_subsidy/apps/subsidy/models.py
johnnagro: @pwnage101 an earlier version of our idea here was to take the geag id into the commit but, we want all committed transactions to be consistent and we may have multiple external fulfillment ids so, we opted to always commit with the enterprise enrollment. make sense?

pwnage101: I'm not sure if you're aware, but Sheehan built up some code and models to allow recording an arbitrary number of "external fulfillment IDs", which this function uses under the hood.  Check out the [ExternalFulfillmentProvider](https://github.com/openedx/openedx-ledger/blob/e340ed60fca5f838b66714659afcf785a07d0f93/openedx_ledger/models.py#L347) and [ExternalTransactionReference](https://github.com/openedx/openedx-ledger/blob/e340ed60fca5f838b66714659afcf785a07d0f93/openedx_ledger/models.py#L388) models from openedx_ledger.

You may first need to create the ExternalFulfillmentProvider records via django admin, then set it in settings so that we know which provider to use from the db, then this code should be creating a new ExternalTransactionReference object and passing it to `commit_transaction(..., external_reference=external_reference)`.  At least, that's the latest plan as I know it.

pwnage101: I was tempted to recommend a data migration to seed the new type (geag allocation id), but I'm recalling this comment: https://github.com/openedx/openedx-ledger/pull/15#discussion_r1153333314 which introduces the concept of a type model specifically to avoid writing code to add a new type.  Guess I'm still on the fence about that because writing code to do this seems less error-prone than manually entering it into django admin then getting the settings right so that we know which record to use.

enterprise_subsidy/apps/fulfillment/api.py
```diff
@@ -32,3 +37,103 @@ def determine_fulfillment_client(subsidy_uuid, content_key):
...
+
+class InvalidFulfillmentMetadataException(FulfillmentException):
+    pass
+
+
+class GEAGFulfillmentHandler():
+    """
+    A class for fulfilling a GEAG transaction.
+    """
+    CENTS_PER_DOLLAR = 100.0
```
adamstankiewicz, line 54: nit/curious: Should we be reusing the existing `CENTS_PER_DOLLAR` constant defined [here](https://github.com/openedx/enterprise-subsidy/blob/30028f60e92bfce584d84513eaf16704299dc149/enterprise_subsidy/apps/subsidy/constants.py#L6)?

enterprise_subsidy/apps/fulfillment/api.py
```diff
@@ -32,3 +37,103 @@ def determine_fulfillment_client(subsidy_uuid, content_key):
...
+    def can_fulfill(self, transaction):
+        """
+        A helper to let callers know if the transaction at hand can be fulfilled
+        with this fulfillment handler.
+        """
+        return self._get_geag_variant_id(transaction)
+
+    def _fulfill_in_geag(self, allocation_payload):
+        geag_response = self.get_smarter_client.create_enterprise_allocation(**allocation_payload)
+        return geag_response.json().get('orderUuid')
```
adamstankiewicz, line 128: [nit/curious/suggestion] Is there any harm in returning the full response JSON here? That is, if/when the GEAG allocation response payload includes more than just the `orderUuid`, we wouldn't have to come back to update this return.

enterprise_subsidy/apps/fulfillment/api.py
```diff
@@ -32,3 +37,103 @@ def determine_fulfillment_client(subsidy_uuid, content_key):
...
+            external_fulfillment_provider=external_fulfillment_provider,
+            external_reference_id=external_reference_id,
+        )
+
+    def can_fulfill(self, transaction):
+        """
+        A helper to let callers know if the transaction at hand can be fulfilled
+        with this fulfillment handler.
+        """
+        return self._get_geag_variant_id(transaction)
```
iloveagent57, line 124: nit: maybe `return bool(self._get_geag_variant_id(transaction))` ?

enterprise_subsidy/apps/fulfillment/tests/test_api.py
```diff
@@ -0,0 +1,202 @@
...
+        transaction = TransactionFactory.create(
+            state=TransactionStateChoices.PENDING,
+            quantity=-19998,
+            ledger=self.subsidy.ledger,
+            lms_user_id=self.lms_user_id,
+            content_key=self.content_key
+        )
+        external_reference_id = uuid4()
+        # pylint: disable=protected-access
+        external_ransaction_reference = self.geag_fulfillment_handler._save_fulfillment_reference(
```
iloveagent57, line 91: missing 't' in 'ransaction'

enterprise_subsidy/apps/fulfillment/tests/test_api.py
```diff
@@ -0,0 +1,202 @@
...
+            transaction,
+            external_reference_id
+        )
+        assert external_ransaction_reference
+        assert external_ransaction_reference.external_reference_id == external_reference_id
```
iloveagent57, line 96: nit: maybe assert something about the provider, too?

enterprise_subsidy/apps/fulfillment/tests/test_api.py
```diff
@@ -0,0 +1,202 @@
...
+            state=TransactionStateChoices.PENDING,
+            quantity=-19998,
+            ledger=self.subsidy.ledger,
+            lms_user_id=self.lms_user_id,
+            content_key=self.content_key,
+            metadata=tx_metadata,
+        )
+        # pylint: disable=protected-access
+        geag_payload = self.geag_fulfillment_handler._create_allocation_payload(transaction)
+        assert geag_payload.get('order_items')[0].get('productId') == content_summary.get('geag_variant_id')
```
iloveagent57, line 128: nit: we should assert that the rest of `geag_payload` has the shape and value we expect, especially the metadata fields.

enterprise_subsidy/apps/fulfillment/tests/test_api.py
```diff
@@ -0,0 +1,202 @@
...
+            state=TransactionStateChoices.PENDING,
+            quantity=-19998,
+            ledger=self.subsidy.ledger,
+            lms_user_id=self.lms_user_id,
+            content_key=self.content_key,
+            metadata=tx_metadata,
+        )
+        with pytest.raises(InvalidFulfillmentMetadataException):
+            # pylint: disable=protected-access
+            self.geag_fulfillment_handler._validate(transaction)
```
iloveagent57, line 170: we should probably check that the omission of any of the required fields raises an exception.  So like, add `geag_terms_accepted_at` to `tx_metadata` and do something like:
```python
for field in the_required_fields:
    transaction.metadata = {key, val for key, val in tx_metadata if key != field}
    with pytest.raises(...):
        self.geag_fulfillment_handler._validate(transaction)
```

enterprise_subsidy/apps/fulfillment/tests/test_api.py
```diff
@@ -0,0 +1,202 @@
...
+
+    @mock.patch("enterprise_subsidy.apps.content_metadata.api.ContentMetadataApi.get_content_summary")
+    @mock.patch("enterprise_subsidy.apps.fulfillment.api.GEAGFulfillmentHandler._fulfill_in_geag")
+    def test_fulfill(self, mock_fulfill_in_geag, mock_get_content_summary):
+        """
+        Ensure basic happy path of `_create_allocation_payload`
```
iloveagent57, line 176: nit: docstring pasta

enterprise_subsidy/apps/fulfillment/tests/test_api.py
```diff
@@ -0,0 +1,202 @@
...
+        }
+        transaction = TransactionFactory.create(
+            state=TransactionStateChoices.PENDING,
+            quantity=-19998,
+            ledger=self.subsidy.ledger,
+            lms_user_id=self.lms_user_id,
+            content_key=self.content_key,
+            metadata=tx_metadata,
+        )
+        assert self.geag_fulfillment_handler.fulfill(transaction)
```
iloveagent57, line 204: nit: other than being truthy, assert something about what the result of `fulfill()` actually is.

enterprise_subsidy/apps/subsidy/models.py
iloveagent57, line 395: we probably want a test for this branch

iloveagent57 (approved):
Looks great!  I have several testing nits.

adamstankiewicz (approved):

iloveagent57 (approved):

PR #71, opened 2023-05-03
feat: Add PUT and PATCH endpoint for subsidies

emrosarioa:
```
* PUT /api/v1/subsidies/{uuid}/
* PATCH /api/v1/subsidies/{uuid}/

ent-7081
```
### Description

### Testing instructions

Add some, if applicable

### Merge checklist
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Commits are squashed


enterprise_subsidy/apps/api/v1/serializers.py
```diff
@@ -269,3 +280,29 @@ class SubsidyCreationRequestSerializer(serializers.Serializer):
...
+        model = Subsidy
+        fields = [
+            "title",
+            "enterprise_customer_uuid",
+            "active_datetime",
+            "expiration_datetime",
+            "unit",
+            "reference_id",
+            "reference_type",
+            "current_balance",
```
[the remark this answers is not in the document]
iloveagent57: Actually I think `current_balance` is defined via a `SerializerMethodField` somewhere above - this field probably won't get returned unless you copy that field definition and method into this class.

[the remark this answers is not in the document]
pwnage101: agreed, you can just drop this field.

enterprise_subsidy/apps/api/v1/tests/test_views.py
```diff
@@ -259,46 +260,40 @@ def test_create_new_subsidy_with_permissions(self, role, status_code):
...
-        data = {
-            "reference_id": "aksdkjtkwekwl88890",
-            "default_title": "title",
-            "default_enterprise_customer_uuid": str(uuid.uuid4()),
-            "default_unit": UnitChoices.USD_CENTS,
-            "default_starting_balance": "10000",
-            "default_revenue_category": RevenueCategoryChoices.BULK_ENROLLMENT_PREPAY,
-            "default_internal_only": True,
-        }
+        factory_data = build(dict, FACTORY_CLASS=SubsidyFactory)
```
pwnage101, line 263: Never seen this before, is this different than SubsidyFactory.build()?

pwnage101 (commented):
still mid review but wanted to drop these

enterprise_subsidy/apps/api/v1/tests/test_views.py
emrosarioa: I think they're basically the same, except SubsidyFactory.build().__dict__ will have a "_state" key, and with build(dict, FACTORY_CLASS=SubsidyFactory) I didn't have to take the extra step to remove "_state"

enterprise_subsidy/apps/api/v1/views/subsidy.py
```diff
@@ -90,6 +90,20 @@ def handle_exception(self, exc):
...
 
+    def get_serializer_class(self, *args, **kwargs):
+        """
+        Return the serializer class to use for the current request.
+
+        We override the function here instead of just setting the ``serializer_class`` class attribute because that
+        attribute only supports using a single serializer for the entire viewset.  This override logic allows for the
+        serializer class to be based conditionally on the type of action.
+        """
+        if self.request.method.lower() in ('put', 'patch'):
```
pwnage101, line 101: I think there's a way to get the current "action", like self.request.action or something.  that should more closely map to the method name being called.

enterprise_subsidy/apps/api/v1/views/subsidy.py
```diff
@@ -272,3 +307,44 @@ def destroy(self, request, *args, **kwargs):
...
+    )
+    def update(self, request, *args, **kwargs):
+        """
+        Update a subsidy
+
+        Endpoint Location: PUT /api/v1/subsidies/{uuid}/
+        """
+        response = super().update(request, *args, **kwargs)
+        instance = self.get_object()
+        response_serializer = SubsidySerializer(instance, context=self.get_serializer_context())
```
iloveagent57, line 329: ah, so this brings up that thing we discussed the other day about overriding `SubsidyUpdateRequestSerializer.to_representation()` - you can have that be something like:
```python
def to_representation(self, instance):
    return SubsidySerializer(instance)
```
That should let you change this method body back to `return super().update(request, *args, **kwargs)`.

iloveagent57: Note: this suggestion is maybe more a stylistic thing - the drawback to it is that the serialization of the response becomes a little less obvious ("I have to look at the request serializer to understand the _response_ serialization?").

enterprise_subsidy/apps/api/v1/views/subsidy.py
```diff
@@ -74,6 +78,32 @@ class SubsidyViewSet(
...
+    def get_serializer_class(self, *args, **kwargs):
+        """
+        Return the serializer class to use for the current request.
+
+        We override the function here instead of just setting the ``serializer_class`` class attribute because that
+        attribute only supports using a single serializer for the entire viewset.  This override logic allows for the
+        serializer class to be based conditionally on the type of action.
+        """
+        if self.request_action.lower() in ('put', 'patch'):
+            return SubsidyUpdateRequestSerializer
```
iloveagent57, line 102: It's odd that this line of code isn't covered by tests...

iloveagent57: I would double-check in the debugger that this line of code is actually executed when running tests.

emrosarioa: You're right, the breakpoint showed that this code wasn't running! The breakpoint worked if I change it to `self.request.method.lower()`.

enterprise_subsidy/apps/api/v1/views/subsidy.py
emrosarioa: I agree, this way is easier to read!

iloveagent57 (approved):
awesome!

PR #80, opened 2023-05-05
docs: subsidy create endpoint parameters are well documented

pwnage101:
This not only adds missing help_texts for each request parameter, but also lists out the acceptible options for each of the enum/choices parameters such as `default_unit` and `default_revenue_category`.


enterprise_subsidy/apps/api/v1/serializers.py
```diff
@@ -256,23 +257,38 @@ class SubsidyCreationRequestSerializer(serializers.Serializer):
...
         required=True,
+        help_text="The positive balance this Subidy will be initially provisioned to start with.",
     )
-    default_revenue_category = serializers.CharField(
+    default_revenue_category = serializers.ChoiceField(
+        RevenueCategoryChoices.CHOICES,
         required=True,
+        help_text=(
+            'Control how revenue is recognized for subsidized enrollments.  In spirit, this is equivalent to the '
+            '"Cataloge Category" for Coupons.  This field is only used downstream analytics and does not change any '
```
emrosarioa, line 287:
```suggestion
            '"Catalog Category" for Coupons.  This field is only used downstream analytics and does not change any '
```

emrosarioa (approved):
LGTM, after the import order formatting issue is fixed! Thank you so much for defining these, it helps me a ton!

enterprise_subsidy/apps/api/v1/serializers.py
pwnage101: This actually is correct---the original app name uses this spelling!

emrosarioa: Good to know!

PR #86, opened 2023-05-08
feat: learner initiated unenrollment monitoring

alex-sheehan-edx:
### Description
learner initiated unenrollment monitoring

### Merge checklist
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Commits are squashed

### DIFF
--- a/docs/decisions/0007-handling-learner-initiated-unenrollments.rst
+++ b/docs/decisions/0007-handling-learner-initiated-unenrollments.rst
@@ -1,0 +1,101 @@
+0007 handling learner initiated unenrollments
+#############################################
+
+Status
+******
+
+**Accepted** May 2023
+
+Context
+*******
+
+The enterprise-subsidy service needs to maintain records of transaction reversals for any associated platform
+fulfillment cancellation. These reversals represent credit, up to the amount allocated by the original transaction,
+returned to the learner's customer's ledger. As there are many sources of a learner initiated unenrollments/
+cancellations, the enterprise subsidy service must take on the responsibility of monitoring the state of enterprise
+fulfillments. This is necessary to ensure that any learner initiated unenrollment or cancellation of an enterprise 
+fulfillment object will be evaluated and that corresponding reversals are written if/when applicable. 
+
+Decision
+********
+
+In a pre-event bus driven architectural world, the enterprise subsidy service will monitor the state of recently 
+unenrolled enterprise fulfillment objects. It will do so by running a management command which will hit the platform
+enterprise fulfillment `API <https://github.com/openedx/edx-enterprise/blob/master/enterprise/api/v1/views.py#L576>`_
+somewhere between once and twice a day. The view will return a list of enterprise fulfillment objects which have been
+unenrolled after the provided datetime query param filter. Notably, the subsidy service will provide an
+unenrolled_after window that is larger than the frequency at which the management command runs. Meaning that if the
+unenrolled_after param is set to 24 hours, the job will run >1 time within a 24 hour period. With these records, the
+enterprise subsidy service will:
+- First check for the existence of a transaction reversal associated with the fulfillment object's transaction and 
+  idempotency key. This is to prevent multiple reversal to be written for the same cancellation action. A learner
+  initiated unenrollment reversal's idempotency key will be formatted as 
+  `unenrollment-reversal-<enterprise_fulfillment_id>-<enterprise_enrollment_unenrolled_at>`.
+
+- Secondly, it will then evaluate whether or not the object should be refunded. Note that the refundability of an 
+  object is not detailed within this ADR and is ultimately up to the provider of the content and the entitlement policy
+  of the course. However if the item is refundable, the subsidy service will then write a reversal record for the
+  transaction.
+
+Consequences    
+************
+
+Event bus driven architecture vs routine jobs and active monitoring
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+
+This management command oriented approach is not ideal as it is not event bus driven. However, the evaluation work done
+on the part of the subsidy service can easily be bootstrapped by future event bus driven implementations. The hope is
+that future iterations on the enterprise subsidy service will be able to hook up an event receiver to the same methods
+run by the management command, but instead of manually fetching and evaluating recently unenrolled records, it would 
+evaluate objects which have triggered an enterprise fulfillment unenrollment/cancellation event. 
+
+Introducing the subsidy service to Jenkins
+++++++++++++++++++++++++++++++++++++++++++
+
+As it currently stands (June 2021), the enterprise subsidy service is not a part of the Jenkins CI pipeline. The
+accepted implementation of running routine jobs for our services hosted on kubernetes is to utilize argoCD. within
+our internal configurations we can ensure that the job is run on a schedule of our choosing.
+
+Admin initiated unenrollment DAG and introducing a transaction reversal writing task
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+
+The subsidy service currently utilizes specific django-signal driven architecture to handle admin initiated 
+unenrollments. This is done by listening for reversal object creations and then making a call to platform to unenroll
+all associated fulfillment objects. This runs in opposition to the planned management command which will write 
+reversals based on the cancellation of fulfillment objects. In order to mitigate any cyclical behaviors between the two
+systems, we will need to implement two stop-gaps.
+
+- The platform enterprise fulfillment cancellation endpoint needs to verify if the fulfillment object is already
+  revoked and ensure that the unenrolled_at time of the course enrollment object is not bumped if so.
+
+- The subsidy service must check if a recently unenrolled fulfillment object already has an associated reversal object.
+
+How the systems will interact with one another
+++++++++++++++++++++++++++++++++++++++++++++++
+
+Records updated and monitored are limited to the subsidy and edx platform services. Notably, platform course enrollment
+and enterprise fulfillment objects are both housed within edx platform but are split between the enterprise package and
+the LMS. The subsidy service will be responsible for maintaining reversals, as well as their transaction counterparts.
+
+The flow for learner initiated unenrollments will be as follows:
+
+1. Learners unenroll from a course. Notably this has multiple origin sources as it can be done from the LMS or from
+   the enterprise learner/admin dashboards, however the end result is that the student.CourseEnrollment object on 
+   platform is de-activated.
+2. Enterprise enrollment and fulfillment objects will be synchronously updated to reflect the unenrollment of the 
+   course enrollment via a Django signal event/handler. 
+3. The subsidy service will routinely hit the enterprise fulfillment recent unenrollment endpoint which will surface
+   enterprise fulfillments tied to recently canceled enterprise enrollments.
+4. The subsidy service will then check for the existence of a reversal object associated with the fulfillment object's
+   transactions. It will then determine if transaction is refundable and if so, write a reversal object. 
+5. After the generation of a transaction reversal, the subsidy service sends a request to platform to unenroll 
+   fulfillment objects which will result in a NOOP as the fulfillments will have already been cancelled.
+
+
+Rejected Alternatives
+*********************
+
+- Instead of running a routine management command, we could rely on event driven architecture to handle the monitoring
+  of fulfillment objects. This would be an ideal implementation, however as the subsidy service is not yet currently
+  hooked up to the event bus, and as there is no prior work done to publish any enterprise related events to the platform
+  event bus, this would require more work to implement than we're willing to invest at the time of this writing.
--- a/enterprise_subsidy/apps/api_client/enterprise.py
+++ b/enterprise_subsidy/apps/api_client/enterprise.py
@@ -3,11 +3,13 @@ Enterprise api client for the subsidy service.
 """
 import logging
 import os
+from datetime import timedelta
 
 import requests
 from django.conf import settings
 
 from enterprise_subsidy.apps.api_client.base_oauth import BaseOAuthClient
+from enterprise_subsidy.apps.core.utils import localized_utcnow
 
 logger = logging.getLogger(__name__)
 
@@ -52,6 +54,12 @@ class EnterpriseApiClient(BaseOAuthClient):
         return os.path.join(
             self.enterprise_fulfillment_url(enterprise_fulfillment_uuid),
             "cancel-fulfillment",
+        )
+
+    def enterprise_fulfillment_unenrollments_url(self):
+        return os.path.join(
+            self.api_base_url,
+            "operator/enterprise-subsidy-fulfillment/unenrolled/",
         )
 
     def get_enterprise_customer_data(self, enterprise_customer_uuid):
@@ -114,13 +122,13 @@ class EnterpriseApiClient(BaseOAuthClient):
         """
         Calls the Enterprise Bulk Enrollment API to enroll learners in courses.
 
-        Arguemnts:
+        Arguments:
             enterprise_customer_uuid (UUID): UUID representation of the customer that the enrollment will be linked to
             enrollment_info (list[dicts]): List of enrollment information required to enroll.
                 Each index must contain key/value pairs:
                     user_id: ID of the learner to be enrolled
                     course_run_key: the course run key to be enrolled in by the user
-                    transaction_id: uuid represenation of the transaction for the enrollment
+                    transaction_id: uuid representation of the transaction for the enrollment
 
                 Example::
                     [
@@ -176,5 +184,24 @@ class EnterpriseApiClient(BaseOAuthClient):
             logger.error(
                 f'Failed to cancel enterprise enrollment for enterprise_fulfillment_uuid: '
                 f'{enterprise_fulfillment_uuid}. Failed with error: {exc}'
+            )
+            raise exc
+
+    def fetch_recent_unenrollments(self):
+        """
+        Fetches enterprise enrollment objects that have been unenrolled within the last 24 hours.
+        """
+        unenrolled_subsidies_url = self.enterprise_fulfillment_unenrollments_url()
+        unenrolled_cutoff = (localized_utcnow() - timedelta(hours=24 * 7)).isoformat()
+        try:
+            response = self.client.get(
+                unenrolled_subsidies_url,
+                params={"unenrolled_after": unenrolled_cutoff}
+            )
+            response.raise_for_status()
+            return response.json()
+        except requests.exceptions.RequestException as exc:
+            logger.error(
+                f'Failed to fetch recently unenrolled enterprise subsidies. Failed with error: {exc}'
             )
             raise exc
--- a/enterprise_subsidy/apps/transaction/management/commands/write_reversals_from_enterprise_unenrollments.py
+++ b/enterprise_subsidy/apps/transaction/management/commands/write_reversals_from_enterprise_unenrollments.py
@@ -1,0 +1,228 @@
+"""
+Management command to fetch enterprise enrollment objects that have been unenrolled within the last 24 hours and write
+Transaction Reversals where appropriate.
+"""
+import logging
+from datetime import datetime, timedelta
+
+from django.contrib import auth
+from django.core.management.base import BaseCommand
+from openedx_ledger.api import reverse_full_transaction
+from openedx_ledger.models import Transaction
+
+from enterprise_subsidy.apps.api_client.enterprise import EnterpriseApiClient
+from enterprise_subsidy.apps.content_metadata.api import ContentMetadataApi
+from enterprise_subsidy.apps.subsidy.models import Subsidy
+from enterprise_subsidy.apps.transaction.utils import generate_transaction_reversal_idempotency_key
+
+logger = logging.getLogger(__name__)
+User = auth.get_user_model()
+
+
+class Command(BaseCommand):
+    """
+    Management command for writing Transaction Reversals from recent Enterprise unenrollments data.
+
+    ./manage.py write_reversals_from_enterprise_unenrollments
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.dry_run_prefix = ""
+        self.dry_run = False
+        self.fetched_content_metadata = {}
+
+    def add_arguments(self, parser):
+        """
+        Entry point for subclassed commands to add custom arguments.
+        """
+        parser.add_argument(
+            '--dry-run',
+            action='store_true',
+            dest='dry_run',
+            help=(
+                'If set, no updates or creates will occur; will instead iterate over '
+                'the unenrollments and log the actions that would have been taken.'
+            ),
+        )
+
+    def unenrollment_can_be_refunded(
+        self,
+        content_metadata,
+        enterprise_course_enrollment,
+    ):
+        """
+        helper method to determine if an unenrollment is refundable
+        """
+        # Retrieve the course start date from the content metadata
+        enrollment_course_run_key = enterprise_course_enrollment.get("course_id")
+        enrollment_created_at = enterprise_course_enrollment.get("created")
+        course_start_date = None
+        for run in content_metadata.get('course_runs'):
+            if run.get('key') == enrollment_course_run_key:
+                course_start_date = run.get('start')
+                break
+
+        if not course_start_date:
+            logger.warning(
+                f"No course start date found for course run: {enrollment_course_run_key}. "
+                "Unable to determine refundability."
+            )
+            return False
+
+        # https://2u-internal.atlassian.net/browse/ENT-6825
+        # OCM course refundability is defined as True IFF:
+        # ie MAX(enterprise enrollment created at, course start date) + 14 days > unenrolled_at date
+        enrollment_created_at = enterprise_course_enrollment.get("created")
+        enrollment_unenrolled_at = enterprise_course_enrollment.get("unenrolled_at")
+        enrollment_created_datetime = datetime.strptime(
+            enrollment_created_at, "%Y-%m-%dT%H:%M:%SZ"
+        )
+        course_start_datetime = datetime.strptime(
+            course_start_date, "%Y-%m-%dT%H:%M:%SZ"
+        )
+        enrollment_unenrolled_at_datetime = datetime.strptime(
+            enrollment_unenrolled_at, "%Y-%m-%dT%H:%M:%SZ"
+        )
+        refund_cutoff_date = max(course_start_datetime, enrollment_created_datetime) + timedelta(days=14)
+
+        if refund_cutoff_date > enrollment_unenrolled_at_datetime:
+            logger.info(
+                f"Course run: {enrollment_course_run_key} is refundable for enterprise customer user: "
+                f"{enterprise_course_enrollment.get('enterprise_customer_user')}. Writing Reversal record."
+            )
+            return True
+        else:
+            logger.info(
+                f"Unenrollment from course: {enrollment_course_run_key} by user: "
+                f"{enterprise_course_enrollment.get('enterprise_customer_user')} is not refundable."
+            )
+            return False
+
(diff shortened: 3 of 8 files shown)


docs/decisions/0006-handling-learner-initiated-unenrollments.rst
```diff
@@ -0,0 +1,100 @@
...
+modified, unenrolled enterprise fulfillment objects. It will do so by running a management command which
+will hit the platform enterprise fulfillment `API 
+<https://github.com/openedx/edx-enterprise/blob/master/enterprise/api/v1/views.py#L576>`_ somewhere between once and
+twice a day. The view will return a list of enterprise fulfillment objects which have been cancelled after the provided
+modified at timestamp filter. Notably, we provide a modified at time window that is smaller than the frequency at which
+the job runs. With these records, the enterprise subsidy service will 
+- First check for the existence of a transaction reversal associated with the fulfillment object's transaction and 
+  idempotency key. This is to prevent multiple reversal to be written for the same cancellation action. A learner
+  initiated unenrollment reversal's idempotency key will be formatted as 
+  `unenrollment-reversal-<enterprise_fulfillment_id>-<enterprise_enrollment_modified>`.
```
pwnage101, line 32: This key won't fly in the case of a learner that redeems, unenrolls, redeems again, and unenrolls again, since the enterprise fulfillment UUID and enterprise enrollment ID with both be the same for both unenrollment attempts.  You probably have to incorporate the transaction UUID too.  But then if a reversal enters "failed" state, re-attempts to unenroll will produce idempotency_key collisions again, so you probably also have to _version_ the idempotency_key for every additional historical failed reversal.

pwnage101: I don't necessarily think we should block on this since I'm not sure it's a v1 thing, but might be worth calling out in the ADR.

pwnage101: Ah, didn't realize `<enterprise_enrollment_modified>` was a date.  That's probably fine then!

docs/decisions/0006-handling-learner-initiated-unenrollments.rst
```diff
@@ -0,0 +1,100 @@
...
+
+Decision
+********
+
+In a pre-event bus driven architectural world, the enterprise subsidy service will monitor the state of recently 
+modified, unenrolled enterprise fulfillment objects. It will do so by running a management command which
+will hit the platform enterprise fulfillment `API 
+<https://github.com/openedx/edx-enterprise/blob/master/enterprise/api/v1/views.py#L576>`_ somewhere between once and
+twice a day. The view will return a list of enterprise fulfillment objects which have been cancelled after the provided
+modified at timestamp filter. Notably, we provide a modified at time window that is smaller than the frequency at which
```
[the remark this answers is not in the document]
pwnage101: I did just notice the timedelta in the code is set to 24 hours, but still, jenkins or whatever runs the management command probably can't guarantee a consistent schedule and could result in a 25+ hour period between runs.  Shouldn't we update the window to something like 7 days to give us some flexibility to recover a stuck job?

enterprise_subsidy/apps/api_client/enterprise.py
```diff
@@ -178,3 +185,20 @@ def cancel_fulfillment(self, enterprise_fulfillment_uuid):
                 f'{enterprise_fulfillment_uuid}. Failed with error: {exc}'
             )
             raise exc
+
+    def fetch_recent_unenrollments(self):
+        """
+        Fetches enterprise enrollment objects that have been unenrolled within the last 24 hours.
+        """
+        unenrolled_subsidies_url = self.enterprise_fulfillment_unenrollment_url()
+        query_param_filter = f'?unenrolled_after={datetime.now() - timedelta(hours=24)}'
```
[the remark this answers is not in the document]
pwnage101: Unrelated nit: consider instead making the client library do the query string encoding:
```python
self.client.get(
    recent_unenrolled_subsidies_url,
    params={"unenrolled_after": unenrolled_cutoff},
)
```

enterprise_subsidy/apps/transaction/management/commands/write_reversals_from_enterprise_unenrollments.py
```diff
@@ -0,0 +1,182 @@
...
+                    if self.determine_refundability_of_unenrollment(content_metadata, enterprise_course_enrollment):
+                        logger.info(
+                            f"Course run: {enrollment_course_run_key} is refundable for enterprise customer user: "
+                            f"{enterprise_course_enrollment.get('enterprise_customer_user')}. Writing Reversal record."
+                        )
+
+                        # Do we need to write any additional metadata to the Reversal?
+                        reverse_full_transaction(
+                            transaction=related_transaction,
+                            idempotency_key=f'unenrollment-reversal-{fulfillment_uuid}-{enrollment_unenrolled_at}',
```
pwnage101, line 170: This idempotency_key is constructed in two distinct places.  Probably could use a helper function to make DRY.

enterprise_subsidy/apps/transaction/management/commands/write_reversals_from_enterprise_unenrollments.py
```diff
@@ -0,0 +1,182 @@
...
+
+        reversals_processed = 0
+        enterprise_catalog_api_client = EnterpriseCatalogApiClient()
+        fetched_content_metadata = {}
+        for unenrollment in recent_unenrollments:
+            fulfillment_uuid = unenrollment.get("uuid")
+            enterprise_course_enrollment = unenrollment.get("enterprise_course_enrollment")
+            enrollment_course_run_key = enterprise_course_enrollment.get("course_id")
+            enrollment_unenrolled_at = enterprise_course_enrollment.get("unenrolled_at")
+            try:
```
pwnage101, line 117: Nit: can you add an `else` block for this `try` to log a message like "Reversal found for unenrollment of enterprise fulfillment: {fulfillment_uuid}. Skipping."

enterprise_subsidy/apps/transaction/management/commands/write_reversals_from_enterprise_unenrollments.py
```diff
@@ -0,0 +1,182 @@
...
+                            idempotency_key=f'unenrollment-reversal-{fulfillment_uuid}-{enrollment_unenrolled_at}',
+                        )
+                        reversals_processed += 1
+                    else:
+                        logger.info(
+                            f"Unenrollment from course: {enrollment_course_run_key} by user: "
+                            f"{enterprise_course_enrollment.get('enterprise_customer_user')} is not refundable."
+                        )
+                else:
+                    logger.info("Found unenrollment related to an externally referenced transaction. Skipping.")
```
pwnage101, line 179: Nit: for completeness, this log message should probably contain the `fulfillment_uuid` since every other branch does already.

enterprise_subsidy/apps/transaction/tests/test_management.py
```diff
@@ -0,0 +1,392 @@
...
+        enrollment_created_at,
+        mock_oauth_client,
+    ):
+        """
+        Test that for write_reversals_from_enterprise_unenrollments, if the greater date between the course start date
+        and the enrollment created at date is more than 14 days before the unenrollment date, no reversal is created.
+        """
+        # unenrolled_at is 14 days after the considered refund period start date so the reversal is not created
+        unenrolled_at = '2023-06-16T19:27:29Z'
+        mock_oauth_client.return_value.get.side_effect = [
```
pwnage101, line 220: This sequencing of API calls feels pretty sensitive to changes.  Would it be possible instead to mock the individual subclasses of the base client, `EnterpriseCatalogApiClient` and `EnterpriseApiClient`?

enterprise_subsidy/apps/api_client/enterprise.py
```diff
@@ -54,6 +55,12 @@ def enterprise_fulfillment_cancel_url(self, enterprise_fulfillment_uuid):
             "cancel-fulfillment",
         )
 
+    def enterprise_fulfillment_unenrollment_url(self):
+        return os.path.join(
+            f"{self.api_base_url}operator/enterprise-subsidy-fulfillment",
+            "unenrolled/",
+        )
```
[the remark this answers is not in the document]
alex-sheehan-edx: yeah this is a relic of iteration, I can just concat the strings 👍

enterprise_subsidy/apps/api_client/enterprise.py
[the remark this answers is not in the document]
alex-sheehan-edx: worth a conversation, I also feel that this is something we can tweak as we go depending on testing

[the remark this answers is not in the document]
alex-sheehan-edx: I was struggling with datetimes a bunch so this is much appreciated! 👍

docs/decisions/0006-handling-learner-initiated-unenrollments.rst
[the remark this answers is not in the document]
alex-sheehan-edx: I apologize, the wording on the ADR was out of date- we have revised our approach here to:

1) pull unenrolled_at instead of modified at on the enterprise enrollment. This is a bit safer of a value because it is _**only**_ written at the time of unenrollment. 

2) The window size (size of the `unenrolled_after` query param used by the subsidy service) of fetched unenrollments will be roughly 2x the pulling frequency (The rate at which the job runs). In this case, the the unenrolled_after would be 24 hours and the job frequency would be set to ~12 hours. This could of course change and I'm open to a conversation but I do believe that a faster frequency is desirable here as we want refunds to be realized relatively quickly. So the question then I guess is what will the `unenrolled_after` value ge? 🤔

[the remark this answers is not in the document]
pwnage101: I'd still advocate for a 7-day window regardless of polling frequency because if the job scheduler (argocd) gets stuck for some reason, we have ~6 days to un-stuck it before we have to manually intervene to expand the window.  Are we concerned that performance won't be up to snuff for this large of a window?

pwnage101 (approved):
Just nits and the 7 day window proposal.  otherwise LGTM!

enterprise_subsidy/apps/transaction/management/commands/write_reversals_from_enterprise_unenrollments.py
```diff
@@ -0,0 +1,174 @@
...
+
+logger = logging.getLogger(__name__)
+User = auth.get_user_model()
+
+
+class Command(BaseCommand):
+    """
+    Management command for writing Transaction Reversals from recent Enterprise unenrollments data.
+
+    ./manage.py write_reversals_from_enterprise_unenrollments
```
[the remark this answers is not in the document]
alex-sheehan-edx: not a bad idea

docs/decisions/0006-handling-learner-initiated-unenrollments.rst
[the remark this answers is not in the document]
alex-sheehan-edx: sure! Let's start with a 7 day window and we can tweak from there. I don't think that it should cause much of an issue performance wise

pwnage101 (approved):

johnnagro (approved):

docs/decisions/0007-handling-learner-initiated-unenrollments.rst
```diff
@@ -0,0 +1,101 @@
...
+++++++++++++++++++++++++++++++++++++++++++
+
+As it currently stands (May 2021), the enterprise subsidy service is not a part of the Jenkins CI pipeline. The
+accepted implementation of running routine jobs for our services hosted on kubernetes is to utilize argoCD. within
+our internal configurations we can ensure that the job is run on a schedule of our choosing.
+
+Admin initiated unenrollment DAG and introducing a transaction reversal writing task
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+
+The subsidy service currently utilizes specific event driven architecture to handle admin initiated unenrollments.
```
iloveagent57, line 62: nit: it's more django-signals based than event-driven, no?

docs/decisions/0007-handling-learner-initiated-unenrollments.rst
```diff
@@ -0,0 +1,101 @@
...
+++++++++++++++++++++++++++++++++++++++++++++++
+
+Records updated and monitored are limited to the subsidy and edx platform services. Notably, platform course enrollment
+and enterprise fulfillment objects are both housed within edx platform but are split between the enterprise package and
+the LMS. The subsidy service will be responsible for maintaining reversals, as well as their transaction counterparts.
+
+The flow for learner initiated unenrollments will be as follows:
+
+1. Learners unenroll from a course. Notably this has multiple origin sources as it can be done from the LMS or from
+   the enterprise learner/admin dashboards, however the end result is that the course enrollment object on platform is
```
iloveagent57, line 83: I might refer to it as the `student.CourseEnrollment` object to help disambiguate.  Also maybe "de-activated" instead of "cancelled" to jibe with the field name.

docs/decisions/0007-handling-learner-initiated-unenrollments.rst
```diff
@@ -0,0 +1,101 @@
...
+   cancelled.
+2. Enterprise enrollment objects will be synchronously updated to reflect the unenrollment of the course enrollment via
```
iloveagent57, line 85: enrollment _and_ fulfillment records

docs/decisions/0007-handling-learner-initiated-unenrollments.rst
```diff
@@ -0,0 +1,101 @@
...
+   a Django signal event/handler. 
+3. The subsidy service will routinely hit the enterprise fulfillment recent unenrollment endpoint which will surface
+   enterprise fulfillments tied to recently canceled enterprise enrollments.
+4. The subsidy service will then check for the existence of a reversal object associated with the fulfillment object's
+   transactions. It will then determine if transaction is refundable and if so, write a reversal object. 
+5. After the generation of a transaction reversal, the subsidy service sends a request to platform to unenroll 
+   fulfillment objects which will result in a NOOP as the fulfillments will have already been cancelled.
```
iloveagent57, line 92: praise: thank you for writing this out, makes it much easier to reason about the big picture!

docs/decisions/0007-handling-learner-initiated-unenrollments.rst
```diff
@@ -0,0 +1,101 @@
...
+
+
+Rejected Alternatives
+*********************
+
+- Instead of running a routine management command, we could rely on event driven architecture to handle the monitoring
+  of fulfillment objects. This would be an ideal implementation, however as the subsidy service is not yet currently
+  hooked up to the event bus, and as there is no prior work done to publish any enterprise related events to the platform
+  event bus, this would require more work to implement than we're willing to invest within PLI MVP.
```
iloveagent57, line 101: nit: I would just say "...than we're willing to invest at the time of this writing" so that whoever is reading this in the future doesn't have to know what "PLI" stands for.

enterprise_subsidy/apps/api_client/enterprise.py
```diff
@@ -178,3 +185,22 @@ def cancel_fulfillment(self, enterprise_fulfillment_uuid):
                 f'{enterprise_fulfillment_uuid}. Failed with error: {exc}'
             )
             raise exc
+
+    def fetch_recent_unenrollments(self):
+        """
+        Fetches enterprise enrollment objects that have been unenrolled within the last 24 hours.
+        """
+        unenrolled_subsidies_url = self.enterprise_fulfillment_unenrollments_url()
+        unenrolled_cutoff = (datetime.now(tz=timezone.utc) - timedelta(hours=24 * 7)).isoformat()
```
iloveagent57, line 194: nit: try using `core.utils.localized_utcnow()`

enterprise_subsidy/apps/api_client/enterprise.py
```diff
@@ -178,3 +185,22 @@ def cancel_fulfillment(self, enterprise_fulfillment_uuid):
...
+        try:
+            response = self.client.get(
+                unenrolled_subsidies_url,
+                params={"unenrolled_after": unenrolled_cutoff}
+            )
+            response.raise_for_status()
+            return response.json()
+        except requests.exceptions.HTTPError as exc:
```
iloveagent57, line 202: Nit: consier using the even more general `requests.exceptions.RequestException`, which will also catch things like connection errors.

enterprise_subsidy/apps/transaction/management/commands/write_reversals_from_enterprise_unenrollments.py
```diff
@@ -0,0 +1,209 @@
...
+                    logger.info(
+                        f"{DRY_RUN_LOG_PREFIX}Found unenrolled enterprise fulfillment: {fulfillment_uuid} related to "
+                        f"an externally referenced transaction: {related_transaction.external_reference.first()}. "
+                        "Skipping."
+                    )
+            else:
+                logger.info(
+                    f"{DRY_RUN_LOG_PREFIX}Found existing Reversal: {existing_reversal} for enterprise fulfillment: "
+                    f"{fulfillment_uuid}. Skipping Reversal creation for Transaction: {related_transaction}."
+                )
```
iloveagent57, line 205: There is a lot of nesting to this logic.  Is there any low-hanging-fruit way to simplify or help make it more readable?

enterprise_subsidy/apps/transaction/management/commands/write_reversals_from_enterprise_unenrollments.py
```diff
@@ -0,0 +1,209 @@
...
+            '--dry-run',
+            action='store_true',
+            dest='dry_run',
+            help=(
+                'If set, no updates or creates will occur; will instead iterate over '
+                'the unenrollments and log the actions that would have been taken.s'
+            ),
+        )
+
+    def determine_refundability_of_unenrollment(
```
iloveagent57, line 43: naming nit: maybe `unenrollment_can_be_refunded()`?

enterprise_subsidy/apps/transaction/management/commands/write_reversals_from_enterprise_unenrollments.py
```diff
@@ -0,0 +1,209 @@
...
+            except Transaction.DoesNotExist:
+                logger.info(
+                    f"{DRY_RUN_LOG_PREFIX}No Subsidy Transaction found for enterprise fulfillment: {fulfillment_uuid}"
+                )
+                continue
+            except Reversal.DoesNotExist:
+                logger.info(
+                    f"{DRY_RUN_LOG_PREFIX}No existing Reversal found for enterprise fulfillment: {fulfillment_uuid}. "
+                    f"Writing Reversal for Transaction: {related_transaction}."
+                )
```
iloveagent57, line 145: nit: I would maybe re-organize this code to not do so much of the real work inside your exception handling blocks.  That would allow to early-continue through your loop when you find that there _is_ already a reversal for the transaction.  The `Transaction.get_reversal()` method will give you a reversal or none, safely.

enterprise_subsidy/apps/transaction/management/commands/write_reversals_from_enterprise_unenrollments.py
```diff
@@ -0,0 +1,209 @@
...
+
+                # On initial release we are only supporting learner initiated unenrollments for OCM courses.
+                # OCM courses are identified by the lack of an external_reference on the Transaction object.
+                # Externally referenced transactions can be unenrolled through the Django admin actions related to the
+                # Transaction model.
+                if not related_transaction.external_reference.exists():
```
iloveagent57, line 151: here's another place where an early continue will help reduce this nesting, e.g.
```python
if related_transaction.external_reference.exists():
    # do your logging
    continue
```

enterprise_subsidy/apps/transaction/management/commands/write_reversals_from_enterprise_unenrollments.py
```diff
@@ -0,0 +1,209 @@
...
+                        try:
+                            customer_uuid = related_transaction.ledger.subsidy.enterprise_customer_uuid
+                        except Subsidy.DoesNotExist:
+                            logger.warning(
+                                f"{DRY_RUN_LOG_PREFIX}Unable to determine enterprise customer uuid for transaction: "
+                                f"{related_transaction}"
+                            )
+                            continue
+
+                        content_metadata = enterprise_catalog_api_client.get_content_metadata_for_customer(
```
iloveagent57, line 163: use the `ContentMetadataApi.get_content_metdata()` function here, which will use the cache for you.

enterprise_subsidy/apps/transaction/management/commands/write_reversals_from_enterprise_unenrollments.py
```diff
@@ -0,0 +1,209 @@
...
+                            customer_uuid,
+                            enrollment_course_run_key,
+                        )
+                        fetched_content_metadata[enrollment_course_run_key] = content_metadata
+                    else:
+                        content_metadata = fetched_content_metadata.get(enrollment_course_run_key)
+
+                    if self.determine_refundability_of_unenrollment(content_metadata, enterprise_course_enrollment):
```
iloveagent57, line 171: another place for early continue: 
```python
if not self.unenrollment_can_be_refunded(...):
    # do your logging
    continue
```

iloveagent57 (commented):
Very nice job with explaining the big picture in your ADR!

enterprise_subsidy/apps/transaction/management/commands/write_reversals_from_enterprise_unenrollments.py
alex-sheehan-edx: @iloveagent57 My only concern with `get_reversal` that I see is the lack of an idempotency key consideration. Is that something we're ok with dropping?

enterprise_subsidy/apps/transaction/management/commands/write_reversals_from_enterprise_unenrollments.py
```diff
@@ -0,0 +1,209 @@
...
+        """
+        parser.add_argument(
+            '--dry-run',
+            action='store_true',
+            dest='dry_run',
+            help=(
+                'If set, no updates or creates will occur; will instead iterate over '
+                'the unenrollments and log the actions that would have been taken.s'
+            ),
+        )
```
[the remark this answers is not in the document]
pwnage101: typo: stranded `s` character and the end of the help text.

enterprise_subsidy/apps/transaction/management/commands/write_reversals_from_enterprise_unenrollments.py
pwnage101: There was a time when we were considering the possibility of multiple reversals for one transaction, but that is not the case anymore.  It is no longer clear to me the utility of an idempotency key for finding existing reversals since the reversal you're looking for can always be precisely located by their linked transaction UUID.  In other words, I'm fine with just doing a transaction-based lookup on reversals.

pwnage101: It still begs the question, why have an idempotency key field at all for reversals!  Rocking the boat a little.

enterprise_subsidy/apps/transaction/management/commands/write_reversals_from_enterprise_unenrollments.py
```diff
@@ -0,0 +1,216 @@
...
+            f"{DRY_RUN_LOG_PREFIX}Updating and writing Transaction Reversals from recent Enterprise unenrollments data"
+        )
+        recent_unenrollments = EnterpriseApiClient().fetch_recent_unenrollments()
+        logger.info(
+            f"{DRY_RUN_LOG_PREFIX}Found {len(recent_unenrollments)} recent Enterprise unenrollments"
+        )
+
+        reversals_processed = 0
+        fetched_content_metadata = {}
+        for unenrollment in recent_unenrollments:
```
johnnagro, line 116: might be nice to take the body of this `for` loop and make a method out of it, like `handle_unenrollment(unenrollment)`

enterprise_subsidy/apps/transaction/management/commands/write_reversals_from_enterprise_unenrollments.py
```diff
@@ -0,0 +1,216 @@
...
+                )
+                continue
+
+            idempotency_key = generate_transaction_reversal_idempotency_key(
+                fulfillment_uuid,
+                enrollment_unenrolled_at
+            )
+
+            # Look for a Reversal related to the unenrollment
+            existing_reversal = Reversal.objects.filter(
```
johnnagro, line 138: you shouldn't filter by idempotency key here - as any valid reversal should be enough to skip this part - maybe `Transaction.get_reversal()` here?

enterprise_subsidy/apps/transaction/management/commands/write_reversals_from_enterprise_unenrollments.py
```diff
@@ -0,0 +1,216 @@
...
+                )
+                continue
+
+            logger.info(
+                f"{DRY_RUN_LOG_PREFIX}Course run: {enrollment_course_run_key} is refundable for enterprise "
+                f"customer user: {enterprise_course_enrollment.get('enterprise_customer_user')}. Writing "
+                "Reversal record."
+            )
+
+            if not dry_run:
```
johnnagro, line 200: you may be able to move the idempotency key generation into this block, to keep it closer to where you're actually going to use it. this co-location makes it easier to read/understand.

enterprise_subsidy/apps/transaction/management/commands/write_reversals_from_enterprise_unenrollments.py
johnnagro: to check for a previous reversal, we shouldn't use the idemptoency key - any prior reversal should count as "this transaction has been reversed". when applying our reversal, we should include our idemptotency key. this is good practice. it prevents multiple copies of the same code from stomping on each other. makes it safer to-rerun.

enterprise_subsidy/apps/transaction/management/commands/write_reversals_from_enterprise_unenrollments.py
johnnagro: This can also make the code easier to test

enterprise_subsidy/apps/transaction/management/commands/write_reversals_from_enterprise_unenrollments.py
johnnagro: idempotency keys are important for callers to supply when taking an action that alters the state of the ledger. idempotency keys are only important on filtering if you really, truely only want to know if this specific code path has already run. usually knowing that _any code path has already taken this action_ is the more correct thing to ask (ie "has this transaction been reversed?" means any code path, "do you have a reversal with this idempotency key?" means this one code path)

johnnagro (approved):

enterprise_subsidy/apps/transaction/utils.py
```diff
@@ -0,0 +1,11 @@
...
+Utility functions used in the implementation of subsidy Transactions.
+"""
+
+
+def generate_transaction_reversal_idempotency_key(fulfillment_uuid, enrollment_unenrolled_at):
+    """
+    Generates a unique idempotency key for a transaction reversal using the fulfillment uuid and time at which the
+    unenrollment occurred.
+    """
+    return f'unenrollment-reversal-{fulfillment_uuid}-{enrollment_unenrolled_at}'
```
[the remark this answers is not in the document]
johnnagro: we really want the `enrollment_unenrolled_at` in here, imho its okay for us to roll a unique method here vs re-using a ledger one

enterprise_subsidy/apps/transaction/management/commands/write_reversals_from_enterprise_unenrollments.py
alex-sheehan-edx: Done!

johnnagro (approved):

PR #87, opened 2023-05-09
feat: include auth_org_id in geag allocation

johnnagro:
### Description

- grab enterprise customer data, extract `auth_org_id`
- pipe `auth_org_id` into geag enterprise allocation call
- profit
- https://2u-internal.atlassian.net/browse/ENT-7079

### DIFF
--- a/enterprise_subsidy/apps/fulfillment/api.py
+++ b/enterprise_subsidy/apps/fulfillment/api.py
@@ -70,6 +70,9 @@ class GEAGFulfillmentHandler():
             api_url=settings.GET_SMARTER_API_URL
         )
 
+    def get_enterprise_client(self, transaction):
+        return transaction.ledger.subsidy.enterprise_client
+
     def _get_geag_transaction_price(self, transaction):
         """
         Get the price in dollars to send to GEAG from transaction quantities,
@@ -83,6 +86,12 @@ class GEAGFulfillmentHandler():
     def _get_geag_variant_id(self, transaction):
         ent_uuid = self._get_enterprise_customer_uuid(transaction)
         return ContentMetadataApi().get_geag_variant_id(ent_uuid, transaction.content_key)
+
+    def _get_auth_org_id(self, transaction):
+        enterprise_customer_uuid = str(self._get_enterprise_customer_uuid(transaction))
+        ent_client = self.get_enterprise_client(transaction)
+        ent_data = ent_client.get_enterprise_customer_data(enterprise_customer_uuid)
+        return ent_data.get('auth_org_id')
 
     def _create_allocation_payload(self, transaction, currency='USD'):
         # TODO: come back an un-hack this once GEAG validation is
@@ -108,6 +117,7 @@ class GEAGFulfillmentHandler():
             'date_of_birth': transaction.metadata.get('geag_date_of_birth'),
             'terms_accepted_at': transaction.metadata.get('geag_terms_accepted_at'),
             'data_share_consent': str(transaction.metadata.get('geag_data_share_consent', 'true')).lower(),
+            'org_id': self._get_auth_org_id(transaction),
         }
 
     def _validate(self, transaction):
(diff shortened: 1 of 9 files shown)


enterprise_subsidy/apps/fulfillment/api.py
```diff
@@ -70,6 +70,9 @@ def get_smarter_client(self):
             api_url=settings.GET_SMARTER_API_URL
         )
 
+    def get_enterprise_client(self, transaction):
+        return transaction.ledger.subsidy.enterprise_client
```
johnnagro, line 74: i am thinking that we may want to re-use the subsidy API client getters, so we can better control caching, etc?

enterprise_subsidy/apps/fulfillment/api.py
```diff
@@ -84,6 +87,12 @@ def _get_geag_variant_id(self, transaction):
         ent_uuid = self._get_enterprise_customer_uuid(transaction)
         return ContentMetadataApi().get_geag_variant_id(ent_uuid, transaction.content_key)
```
johnnagro, line 88: if we like the pattern of using the client from the subsidy, we can do a similar treatment here

enterprise_subsidy/apps/fulfillment/api.py
pwnage101: LGTM

[non-blocking aside] We do a lot of caching client getters inside model/viewset instance methods, but have we considered instead caching API keys globally within the client module so that they transparently persist across the whole request?  I feel like the current paradigm of cached client getters inside instances methods have a couple downsides:

* unit testing: can make it difficult to mock a client because there could be multiple different classes within which the cached getter is defined, and they all have to be tracked down and mocked or else the client will not appear consistent, and sometimes the cache could cause bleed between subsequent unit tests (especially if the client module is patched instead of the instance getter, an easy mistake to make).
* every new viewset or model class has to independently define a cached client getter, and potentially even duplicate logic to read settings to get an API version number.

I feel like there must be a good, robust way of designing a client that is self-caching across the whole request, and also is a single point of mocking for unit testing.  Thoughts?

enterprise_subsidy/apps/fulfillment/tests/test_api.py
```diff
@@ -98,7 +98,8 @@ def test_save_reference(self):
         assert this_slug == self.geag_fulfillment_handler.EXTERNAL_FULFILLMENT_PROVIDER_SLUG
 
     @mock.patch("enterprise_subsidy.apps.content_metadata.api.ContentMetadataApi.get_content_summary")
-    def test_create_allocation_payload(self, mock_get_content_summary):
+    @mock.patch("enterprise_subsidy.apps.fulfillment.api.GEAGFulfillmentHandler._get_auth_org_id")
```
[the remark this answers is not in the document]
pwnage101: +1

enterprise_subsidy/apps/fulfillment/api.py
iloveagent57: Yup, an idea I wanted to bring up for discussion tomorrow is "caching at the boundaries of the system", i.e. cache our views/outputs for the sake of clients; cache replicated data (our inputs) right when we read that data.  
I think we've brought up the pattern of a read-through cache before, which should align with this idea.
https://hazelcast.com/blog/a-hitchhikers-guide-to-caching-patterns/

iloveagent57: Another idea to bring up: persist cached data in models rather than in a cache-proper, so that we can introspect the state of cached data more easily, or even have something like a "schema" that describes the shape of the data.
We have one ok-ish pattern for this in https://github.com/openedx/enterprise-catalog/blob/master/enterprise_catalog/apps/api_client/enterprise_cache.py

iloveagent57 (approved):

PR #91, opened 2023-05-13
fix: use product source slug

adamstankiewicz:
### Description

Uses `product_source`'s `slug` instead of `name`, as the `name` field should be assumed to be changed at any time, although name and slug are currently identical on stage/prod.

### Testing instructions

Add some, if applicable

### Merge checklist
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Commits are squashed

### DIFF
--- a/enterprise_subsidy/apps/content_metadata/api.py
+++ b/enterprise_subsidy/apps/content_metadata/api.py
@@ -70,7 +70,7 @@ class ContentMetadataApi:
         Helps get the product source string, given a dict of ``content_data``.
         """
         if product_source := content_data.get('product_source'):
-            source_name = product_source.get('name')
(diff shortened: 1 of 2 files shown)


enterprise_subsidy/apps/content_metadata/api.py
```diff
@@ -70,7 +70,7 @@ def product_source_for_content(self, content_data):
         Helps get the product source string, given a dict of ``content_data``.
         """
         if product_source := content_data.get('product_source'):
-            source_name = product_source.get('name')
+            source_name = product_source.get('slug')
```
adamstankiewicz, line 74: Prefer using the slug over name, as name is usually something that can change at any time.

macdiesel (approved):

PR #101, opened 2023-05-17
feat: use request cache to store content metadata results.

iloveagent57:
### Description
Stores content metadata payload in the request cache, which should mean that we make at most one call per content key to enterprise-catalog from the enterprise-subsidy per `can_redeem` request.

### Testing instructions
* Run `make app-logs | grep enterprise-catalog` to tail the logs
* Make a request to `can_redeem`, either via enterprise-access, or directly against enterprise-subsidy
* Watch for only 1 request to enterprise-catalog
* Do this outside of this branch to notice that there are two such requests.

### Merge checklist
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Commits are squashed

### DIFF
--- a/enterprise_subsidy/apps/content_metadata/api.py
+++ b/enterprise_subsidy/apps/content_metadata/api.py
@@ -5,7 +5,10 @@ during subsidy redemption and fulfillment.
 import logging
 from decimal import Decimal
 
+from edx_django_utils.cache import RequestCache
+
 from enterprise_subsidy.apps.api_client.enterprise_catalog import EnterpriseCatalogApiClient
+from enterprise_subsidy.apps.core.utils import versioned_cache_key
 from enterprise_subsidy.apps.subsidy.constants import CENTS_PER_DOLLAR
 
 from .constants import CourseModes, ProductSources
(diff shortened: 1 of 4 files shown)


enterprise_subsidy/apps/content_metadata/api.py
```diff
@@ -19,6 +22,22 @@
...
+    Returns a versioned cache key that includes the customer uuid and content_key.
+    """
+    return versioned_cache_key(enterprise_customer_uuid, content_key)
+
+
+def request_cache():
+    """
+    Helper that returns a namespaced RequestCache instance.
+    """
+    return RequestCache(namespace=CACHE_NAMESPACE)
```
pwnage101, line 39: RequestCache is thread-local, so do you even need to put the code version in the cache?  Every new instance, let alone new deployment, will use a different cache, right?  Perhaps you meant to use TieredCache instead?

macdiesel (approved):

johnnagro (approved):

enterprise_subsidy/apps/content_metadata/api.py
iloveagent57: We'll eventually used TieredCache, I wanted to get this versioned-cache key down on paper to establish a pattern that makes key-based invalidation the "go-to" in this repo.

PR #106, opened 2023-05-21
fix: ensure new transactions to enroll in previously unenrolled courses flips the enrollment to active

adamstankiewicz:
### Description

tl;dr; When redeeming a policy/subsidy for a course run that had been previously redeemed and then had the associated transaction and enrollment reversed, we don't allow the learner to redeem the course again but treat it as a success. The reason is that we see a committed transaction but don't take into the account the fact that it was reversed. So, we don't attempt redemption again when we should, so the learner's enrollment stays inactive.

#### Current state

1. Redeem a subsidy for some content by clicking the "Enroll" CTA on the course page and get redirected to courseware.
2. Return to course page and see "View course" button indicating existing enrollment.
3. Refund & unenroll the learner.
4. Observe that `CourseEnrollment.is_active=False` and the "Enroll" CTA returns on the course page.
5. Click the "Enroll" CTA and get redirected to courseware, with the expectation my previously unenrolled course is now active.
6. Return to course page and observe that `CourseEnrollment.is_active=False` still and the "Enroll" CTA is still shown, even though I had a successful POST `redeem` to enterprise-access.

#### Expected state

1. Redeem a subsidy for some content by clicking the "Enroll" CTA on the course page and get redirected to courseware.
2. Return to course page and see "View course" CTA indicating existing enrollment.
3. Refund & unenroll the learner.
4. Observe that `CourseEnrollment.is_active=False` and the "Enroll" CTA returns on the course page.
5. Click the "Enroll" CTA and get redirected to courseware, with the expectation my previously unenrolled course is now active.
6. Return to course page and and observe that `CourseEnrollment.is_active=True` now and the "View course" CTA is shown instead.

### Merge checklist
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Commits are squashed

### DIFF
--- a/enterprise_subsidy/apps/subsidy/tests/test_api.py
+++ b/enterprise_subsidy/apps/subsidy/tests/test_api.py
@@ -103,11 +103,11 @@ class CanRedeemTestCase(TestCase):
             lms_user_id=self.lms_user_id,
             content_key=self.content_key
         )
-        reversal = Reversal.objects.create(
+        Reversal.objects.create(
             transaction=existing_transaction,
             idempotency_key=str(existing_transaction.idempotency_key) + '-reversed',
             quantity=19998,
-            state=TransactionStateChoices.COMMITTED,
+            state=TransactionStateChoices.CREATED,
         )
         expected_redeemable = True
         expected_price = 19998
@@ -117,8 +117,7 @@ class CanRedeemTestCase(TestCase):
         )
         self.assertEqual(expected_redeemable, actual_redeemable)
         self.assertEqual(expected_price, actual_price)
-        self.assertEqual(existing_transaction, actual_transaction)
-        self.assertEqual(actual_transaction.reversal, reversal)
+        self.assertEqual(None, actual_transaction)
 
     def test_existing_transaction_no_reversal(self):
         """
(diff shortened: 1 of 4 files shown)


enterprise_subsidy/apps/subsidy/tests/test_api.py
```diff
@@ -103,11 +103,11 @@ def test_existing_transaction_with_reversal(self):
             lms_user_id=self.lms_user_id,
             content_key=self.content_key
         )
-        reversal = Reversal.objects.create(
+        Reversal.objects.create(
             transaction=existing_transaction,
             idempotency_key=str(existing_transaction.idempotency_key) + '-reversed',
             quantity=19998,
-            state=TransactionStateChoices.COMMITTED,
+            state=TransactionStateChoices.CREATED,
```
adamstankiewicz, line 110: Reversal objects seem to be created with `state=CREATED`, so updating this to reflect that reality.

pwnage101: If the steady state for a reversal is state=CREATED, then that is a bug.   I recommend that we leave this as it was, and I can pick up a fix in a separate PR.

pwnage101 (approved):

PR #120, opened 2023-06-01
chore: add is_active field to subsidy API read view

emrosarioa:
### Description
Add `is_active field` to  the `Subsidy` API read view

### Testing instructions

Add some, if applicable

### Merge checklist
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Commits are squashed

### DIFF
--- a/enterprise_subsidy/apps/api/v1/serializers.py
+++ b/enterprise_subsidy/apps/api/v1/serializers.py
@@ -21,6 +21,7 @@ class SubsidySerializer(serializers.ModelSerializer):
     Serializer for the `Subsidy` model.
     """
     current_balance = serializers.SerializerMethodField(help_text="The current (remaining) balance of this subsidy.")
+    is_active = serializers.BooleanField(read_only=True, help_text="Whether this subsidy is currently active.")
 
     class Meta:
         """
@@ -40,6 +41,7 @@ class SubsidySerializer(serializers.ModelSerializer):
             "starting_balance",
             "internal_only",
             "revenue_category",
+            "is_active",
             # In the MVP implementation, there are only learner_credit subsidies.  Uncomment after subscription
             # subsidies are introduced.
             # "subsidy_type",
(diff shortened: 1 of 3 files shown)


enterprise_subsidy/apps/subsidy/models.py
```diff
@@ -204,7 +205,16 @@ def is_active(self):
         Returns true if the localized current time is
         between ``active_datetime`` and ``expiration_datetime``.
         """
-        return self.active_datetime <= localized_utcnow() <= self.expiration_datetime
+        active_datetime = self.active_datetime
+        expiration_datetime = self.expiration_datetime
+
+        if isinstance(self.active_datetime, str):
```
emrosarioa, line 211: This is due to the tests, the `active_datetime` and `expiration_datetime` are created with `factory.LazyFunction(fake_datetime)`. I thought it wouldn't be too problematic to have a type check here, just in case, but I can also change the tests so they aren't using strings.

pwnage101: I'm a fan of structuring production code paths to make testing easier, but this feels a bit too weird.  It seems odd that the model field itself has type `DateTimeField`, and the factory field function returns a `datetime` object, yet when recalled it somehow returns a string?

emrosarioa: I was a bit too hasty blaming the issue on the test itself. After adding some breakpoints, the "date as a string" problem happens when the subsidy is the result of the `get_or_create_learner_credit_subsidy` function. I added `parse_datetime` inside the `create` function, but I'm curious if you feel there's a better spot to make sure the dates aren't strings?

johnnagro (approved):

enterprise_subsidy/apps/subsidy/models.py
emrosarioa: I think I figured it out! Reading the [DRF docs here](https://www.django-rest-framework.org/api-guide/serializers/#deserializing-objects), it says to use `serializer.validated_data` after `serializer.is_valid()`. I made the change in the `SubsidyCreationRequestSerializer` from `serializer.data`, and it solved the string/datetime issue.

pwnage101 (approved):
nice!

PR #126, opened 2023-06-08
fix: Reversal.metadata field is not correctly serialized

pwnage101:
* API before: the output-serialized reversal metadata key pointed to a string value containing JSON
* API after: the output-serialized reversal metadata key pointed to a nested JSON dictionary.

redoc before:
image
redoc after:
image

### DIFF
--- a/enterprise_subsidy/apps/api/v1/serializers.py
+++ b/enterprise_subsidy/apps/api/v1/serializers.py
@@ -62,6 +62,9 @@ class ReversalSerializer(serializers.ModelSerializer):
     """
     Serializer for the `Reversal` model.
     """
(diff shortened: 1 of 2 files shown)


enterprise_subsidy/apps/api/v1/serializers.py
```diff
@@ -78,6 +81,14 @@ class Meta:
             "modified",
         ]
 
+    @extend_schema_field(serializers.JSONField)
+    def get_metadata(self, obj) -> dict:
```
alex-sheehan-edx, line 85: :surprised_pikachu: wow that's a new notation I haven't seen before

alex-sheehan-edx (approved):
:shipit:

PR #135, opened 2023-07-07
chore: only allow course unenrollment if transaction is committed

emrosarioa:
### Description
If a transaction is not in a COMMITTED state, it should not allow unenrolls or reversals. This PR only changes the management system. It would be a corner case for a transaction linked to an enrollment object to be anything but COMMITTED, but this should help if a corner case is encountered.

### Merge checklist
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Commits are squashed

### DIFF
--- a/enterprise_subsidy/apps/transaction/management/commands/write_reversals_from_enterprise_unenrollments.py
+++ b/enterprise_subsidy/apps/transaction/management/commands/write_reversals_from_enterprise_unenrollments.py
@@ -8,7 +8,7 @@ from datetime import datetime, timedelta
 from django.contrib import auth
 from django.core.management.base import BaseCommand
 from openedx_ledger.api import reverse_full_transaction
-from openedx_ledger.models import Transaction
+from openedx_ledger.models import Transaction, TransactionStateChoices
 
 from enterprise_subsidy.apps.api_client.enterprise import EnterpriseApiClient
 from enterprise_subsidy.apps.content_metadata.api import ContentMetadataApi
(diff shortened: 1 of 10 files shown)


macdiesel (approved):

enterprise_subsidy/apps/transaction/management/commands/write_reversals_from_enterprise_unenrollments.py
```diff
@@ -126,6 +126,13 @@ def handle_reversing_enterprise_course_unenrollment(self, unenrollment):
             )
             return 0
 
+        if related_transaction.state != TransactionStateChoices.COMMITTED:
+            logger.info(
+                f"{self.dry_run_prefix}Transaction: {related_transaction} is not in a committed state. "
+                f"Skipping Reversal creation."
+            )
+            return 0
```
pwnage101, line 134: hm, since we're changing the behavior of `reverse_full_transaction()` in the openedx-ledger PR, and that same function is called in this function below, does that make this check somewhat redundant?  I'm generally in favor of keeping it since it causes an early termination, bypassing all the subsequent checks, but maybe add a comment to the code along the lines of "# Fail early if the transaction is not committed, even though reverse_full_transaction() would throw an exception later anyway.".

pwnage101 (approved):

enterprise_subsidy/apps/transaction/management/commands/write_reversals_from_enterprise_unenrollments.py
emrosarioa: Yes that's a great point! I'll add a comment.

macdiesel (approved):

PR #136, opened 2023-07-13
fix: Update subsidy paginator logic to reference response object

brobro10000:
Updates logic of the `SubsidyListPaginator` to correctly reference the correct Subsidy response object data. 
Split logic to validate a query parameter exist for `page_size` when assigning the paginator value. 

### Description

### Testing instructions

Add some, if applicable

### Merge checklist
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Commits are squashed

### DIFF
--- a/enterprise_subsidy/apps/api/paginators.py
+++ b/enterprise_subsidy/apps/api/paginators.py
@@ -55,12 +55,15 @@ class SubsidyListPaginator(pagination.PageNumberPagination):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.page_count = 0
-        self.page_size_query_param = None
+        self.page_count = None
+        self.pagination_page_size = None
 
     def paginate_queryset(self, queryset, request, view=None):
-        self.page_count = Subsidy.objects.count()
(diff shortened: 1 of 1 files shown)


enterprise_subsidy/apps/api/paginators.py
```diff
@@ -55,18 +55,20 @@ class SubsidyListPaginator(pagination.PageNumberPagination):
...
+        self.page_count = None
+        self.pagination_page_size = None
 
     def paginate_queryset(self, queryset, request, view=None):
-        self.page_count = Subsidy.objects.count()
-        self.page_size_query_param = request.query_params.get('page_size') or self.page_size
+        if request.query_params.get('page_size'):
+            self.pagination_page_size = int(request.query_params.get('page_size'))
+        else:
+            self.pagination_page_size = self.page_size
```
iloveagent57, line 65: nit: you can do this as one statement in python with something like
```python
self.pagination_page_size = int(request.query_params.get('page_size', self.page_size))
```
Where `self.page_size` is the default value returned by `.get()` in case the `page_size` key isn't present in the query_params dict.

iloveagent57 (approved):

adamstankiewicz (approved):
LGTM!

PR #138, opened 2023-07-24
feat: add 'active' flag to subsidies, 'can-redeem' endpoint

brobro10000:
Adds an active response to the 'can-redeem' endpoint.

Its purpose is to allow the policies 'can-redeem' endpoint to be able to determine if the subsidy is active and return the correct messaging to the user in the front end

Related to PR:  https://github.com/openedx/enterprise-access/pull/223

### Description

### Testing instructions

Add some, if applicable

### Merge checklist
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Commits are squashed


enterprise_subsidy/apps/subsidy/api.py
```diff
@@ -90,6 +90,7 @@ def can_redeem(subsidy, lms_user_id, content_key):
       )
     """
     existing_transaction = subsidy.get_committed_transaction_no_reversal(lms_user_id, content_key)
+    is_active = subsidy.is_active
```
macdiesel, line 93: Since the is_active field is not part of the model I would have expected you to calculate the is_active flag here and not just assign it? I don't think this will work.

macdiesel: Hamzah set me straight:
> Ahh the is_active field is computed against the expiration date. There was no need to recompute it again, just assign a value based on the subsidy. If it’s true or false return that value

macdiesel (approved):

PR #148, opened 2023-09-06
feature: enable soft-delete in subsidy model

emrosarioa:
### Description
Adds a new field `is_soft_deleted` to the `Subsidy` model, as well as the `ActiveSubsidyManager` to filter out subsidies that are marked `is_soft_deleted=True`.  The ability to delete the subsidy in Django admin is removed.

### Testing instructions

Screenshot 2023-09-11 at 9 13 50 AM

Screenshot 2023-09-11 at 9 13 36 AM

### Merge checklist
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Commits are squashed


enterprise_subsidy/apps/subsidy/migrations/0017_auto_20230905_1511.py
```diff
@@ -0,0 +1,47 @@
...
+        migrations.AlterModelOptions(
+            name='subsidy',
+            options={'ordering': ['-created']},
+        ),
+        migrations.AddField(
+            model_name='historicalsubsidy',
+            name='is_hidden',
+            field=models.BooleanField(default=False, help_text='If set, this subsidy will be considered as soft-deleted or deactivated.'),
+        ),
+        migrations.AddField(
```
johnnagro, line 22: how about using `hidden_at` (a datetime) instead? this lets us know _when_ something was deleted.

iloveagent57: We have history on this model that would let us know that, so either way is more-or-less the same.  I think `is_hidden` is slightly more user-friendly.

macdiesel: Nit: While we are nit picking... why `is_hidden` and not `is_deleted`? For me this falls under something like be explicit in your naming and say what you are actually going to do. Are we hiding here or deleting? I think that deleting makes what is happening more obvious to later readers/users of the code.

iloveagent57: "hidden" is the best name we came up with at the time of grooming and ticket-writing.  `is_deleted` or `deleted_at` both work equally fine, but I think we're bike-shedding.

enterprise_subsidy/apps/subsidy/migrations/0017_auto_20230905_1511.py
```diff
@@ -0,0 +1,47 @@
+# Generated by Django 3.2.20 on 2023-09-05 15:11
```
iloveagent57, line 1: naming nit: we generally try to give the migration files a descriptive name when they're created, using the `-n` flag, e.g. `./manage.py makemigrations subsidy -n add_subsidy_is_hidden`.
This makes it easier to see at a glance which migrations have been applied which running `./manage.py showmigrations subsidy`.

enterprise_subsidy/apps/subsidy/models.py
```diff
@@ -197,8 +208,15 @@ class Meta:
         default=None,
         help_text="The datetime when this Subsidy is considered expired.  If null, this Subsidy is considered active."
         )
+    is_hidden = models.BooleanField(
+        default=False,
+        help_text="If set, this subsidy will be considered as soft-deleted or deactivated.",
+    )
```
iloveagent57, line 214: Since we've included this field in the filter for the default model manager, we'll want to add an index on it to keep all of the queries performant.

enterprise_subsidy/apps/subsidy/tests/test_models.py
```diff
@@ -288,3 +288,28 @@ def test_redeem_with_geag_exception(self, mock_get_content_summary, mock_enterpr
...
+    def test_active_subsidy_manager(self):
+        """
+        Test that the ActiveSubsidyManager only retrieves non-hidden subsidies
+        """
+        SubsidyFactory.create(is_hidden=True)
+        SubsidyFactory.create(is_hidden=False)
+        self.assertEqual(Subsidy.objects.count(), 1)
+        self.assertEqual(Subsidy.all_objects.count(), 2)
+
+    def test_all_objects_manager(self):
```
iloveagent57, line 309: Nit: you basically did this test two lines above :)

enterprise_subsidy/apps/subsidy/migrations/0017_auto_20230905_1511.py
```diff
@@ -0,0 +1,47 @@
...
+        ),
+        migrations.AlterField(
+            model_name='subsidy',
+            name='active_datetime',
+            field=models.DateTimeField(default=None, help_text='The datetime when this Subsidy is considered active.  If null, this Subsidy is considered active.', null=True),
+        ),
+        migrations.AlterField(
+            model_name='subsidy',
+            name='expiration_datetime',
+            field=models.DateTimeField(default=None, help_text='The datetime when this Subsidy is considered expired.  If null, this Subsidy is considered active.', null=True),
```
iloveagent57, line 45: How did these two fields get included in the migration?  I don't see corresponding model changes.

enterprise_subsidy/apps/subsidy/migrations/0017_auto_20230905_1511.py
```diff
@@ -0,0 +1,47 @@
...
+        ),
+        migrations.AlterField(
+            model_name='historicalsubsidy',
+            name='expiration_datetime',
+            field=models.DateTimeField(default=None, help_text='The datetime when this Subsidy is considered expired.  If null, this Subsidy is considered active.', null=True),
+        ),
+        migrations.AlterField(
+            model_name='subsidy',
+            name='active_datetime',
+            field=models.DateTimeField(default=None, help_text='The datetime when this Subsidy is considered active.  If null, this Subsidy is considered active.', null=True),
```
iloveagent57, line 40: ...but also, the existing `help_text` is misleading.  Subsidies with a null `active_datetime` shouldn't be considered active, the code considers them to be _inactive_.  So it'd be good to fix that up when you re-run migrations to add an index on the hidden field.

pwnage101 (commented):
two high level things:

1. Consider the flow of deleting a subsidy, then creating a new one that is identical.  There is some idempotency logic that may claim that the subsidy already exists and happily return a 200 OK.  We should explicitly decide what the behavior should be, and explicitly unit test that delete-create scenario.
2. This is a genuine feature, not just a chore.  Commit message should start with "feat: "

enterprise_subsidy/apps/api/v1/views/subsidy.py
```diff
@@ -326,12 +325,14 @@ def create(self, request, *args, **kwargs):
...
-        Delete a subsidy
+        Soft delete a subsidy by setting its is_soft_deleted field to True
 
         Endpoint Location: DELETE /api/v1/subsidies/{uuid}/
         """
-        response = super().destroy(request, kwargs['uuid'])
-        return response
+        instance = self.get_object()
+        instance.is_soft_deleted = True
+        instance.save()
```
iloveagent57, line 334: It'd be better to override the `delete()` method on the model to do this, so that any other consumer of the model (the Django Admin, management commands, or whatever) will have almost no choice but to soft-delete.
Doing a soft-delete inside the model definition would also let you remove the admin modifications around delete permissions in this PR.

enterprise_subsidy/apps/api/v1/views/subsidy.py
```diff
@@ -326,12 +325,13 @@ def create(self, request, *args, **kwargs):
...
-        Delete a subsidy
+        Soft delete a subsidy by setting its is_soft_deleted field to True
 
         Endpoint Location: DELETE /api/v1/subsidies/{uuid}/
         """
-        response = super().destroy(request, kwargs['uuid'])
-        return response
+        instance = self.get_object()
+        instance.delete()
+        return Response(status=status.HTTP_204_NO_CONTENT)
```
iloveagent57, line 334: Just curious, did you try leaving this method body alone?  I kinda figured that `super().destroy(...)` would basically do the same thing that you're explicitly doing here in this `destroy` view.

enterprise_subsidy/apps/api/v1/tests/test_views.py
```diff
@@ -374,18 +374,12 @@ def test_valid_delete_subsidy_by_role(self, role, status_code):
...
-    def test_delete_subsidy_with_invalid_uuid(self):
-        self.set_up_operator()
-        response = self.client.delete(self.get_details_url([str(uuid.uuid4())]))
-
-        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
-        self.assertDictEqual(response.json(), {
-            "error_code": "Forbidden",
-            "developer_message": "MISSING: subsidy.can_write_subsidies",
-            "user_message": "MISSING: subsidy.can_write_subsidies",
-        })
```
iloveagent57, line 388: Is there a reason for deleting this test?

iloveagent57 (approved):
This LGTM, just two small questions!

enterprise_subsidy/apps/api/v1/tests/test_views.py
emrosarioa: You're right, it still applies, I'll add it back in

PR #166, opened 2023-10-30
feat: skip customer fetch in content_metadata view

iloveagent57:
### Description
Include the `skip_customer_fetch` param on calls to fetch customer content metadata. See https://github.com/openedx/enterprise-catalog/pull/702

### Merge checklist
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Commits are squashed

### DIFF
--- a/enterprise_subsidy/apps/api_client/enterprise_catalog.py
+++ b/enterprise_subsidy/apps/api_client/enterprise_catalog.py
@@ -31,7 +31,9 @@ class EnterpriseCatalogApiClient(BaseOAuthClient):
             f'content-metadata/{content_identifier}/'
         )
 
-    def get_content_metadata_for_customer(self, enterprise_customer_uuid, content_identifier):
+    def get_content_metadata_for_customer(
+        self, enterprise_customer_uuid, content_identifier, skip_customer_fetch=False, **kwargs
+    ):
         """
         Returns Enterprise Customer related data for a specified piece on content.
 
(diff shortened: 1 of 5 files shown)


enterprise_subsidy/apps/api_client/enterprise_catalog.py
```diff
@@ -40,6 +42,8 @@ def get_content_metadata_for_customer(self, enterprise_customer_uuid, content_id
             content_identifier (str): **Either** the content UUID or content key identifier for a content record.
                 Note: the content needs to be owned by a catalog associated with the provided customer else this
                 method will throw an HTTPError.
+            skip_customer_fetch (bool): Forces enterprise-catalog to skip a sub-call to an edx-enterprise
+                API endpoint running in the edx-platform runtime. Defaults to True.
```
adamstankiewicz, line 46: nit: would it be helpful to describe what the purpose of the customer fetch API endpoint is for / why one might want to disable/enable it? Seems like it might be helpful business context to include at the code-level as well.

adamstankiewicz, line 46: [clarification] This comment suggests the default is `True`, but the function argument has the default set as `False`.

adamstankiewicz (approved):
One clarifying question around whether `skip_customer_fetch` is intended to be enabled or disabled by default and a minor docs nit. Looks good, though!

PR #168, opened 2023-10-31
fix: validate subsidy uniqueness with clean() method.

iloveagent57:
Use a clean method to validate that non-internal-only subsidies must be unique on (reference_id, reference_type), since mysql does not support conditional unique constraints. ENT-7891

### Merge checklist
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Commits are squashed

### DIFF
--- a/enterprise_subsidy/apps/subsidy/tests/test_models.py
+++ b/enterprise_subsidy/apps/subsidy/tests/test_models.py
@@ -1,12 +1,14 @@
 """
 Tests for functionality provided in the ``models.py`` module.
 """
+import random
 from itertools import product
 from unittest import mock
 from uuid import uuid4
 
 import ddt
 import pytest
+from django.core.exceptions import ValidationError
 from django.test import TestCase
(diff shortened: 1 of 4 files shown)


enterprise_subsidy/apps/subsidy/tests/test_models.py
```diff
@@ -99,6 +101,49 @@ def test_price_for_content_not_in_catalog(self):
         with self.assertRaises(ContentNotFoundForCustomerException):
             self.subsidy.price_for_content('some-content-key')
 
+    def test_reference_uniqueness(self):
+        """
+        Tests that not soft-deleted, non-internal-only subsidies
```
adamstankiewicz, line 106: nit/clarification: Is it worth having a test case that covers when a subsidy _is_ soft-deleted? At first glance, it's not immediately clear what would happen differently with this logic when a subsidy is soft-deleted, if anything.

PR #181, opened 2023-11-20
feat: backpopulate transaction email and content title

johnnagro:
### Description

- backpopulate `lms_user_email` and `content_title` for `Transactions`
- https://2u-internal.atlassian.net/browse/ENT-7974
- based on work done in https://github.com/openedx/enterprise-subsidy/pull/162

### DIFF
--- a/enterprise_subsidy/apps/transaction/management/commands/backpopulate_transaction_email_and_title.py
+++ b/enterprise_subsidy/apps/transaction/management/commands/backpopulate_transaction_email_and_title.py
@@ -1,0 +1,93 @@
+"""
+Management command to backpopulate transaction email and title
+"""
+import logging
+
+from django.core.management.base import BaseCommand
+from django.db.models import Q
+from openedx_ledger.models import Transaction
+
+from enterprise_subsidy.apps.subsidy.models import Subsidy
+from enterprise_subsidy.apps.transaction.utils import batch_by_pk
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    """
+    Management command for backpopulating transaction email and title
+
+    ./manage.py backpopulate_transaction_email_and_title
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.dry_run = False
+        self.include_internal_subsidies = False
+
+    def add_arguments(self, parser):
+        """
+        Entry point for subclassed commands to add custom arguments.
+        """
+        parser.add_argument(
+            '--dry-run',
+            action='store_true',
+            dest='dry_run',
+            help=(
+                'If set, no updates will occur; will instead log '
+                'the actions that would have been taken.'
+            ),
+        )
+        parser.add_argument(
+            '--include-internal-subsidies',
+            action='store_true',
+            dest='include_internal_subsidies',
+            help=(
+                'If set, internal subsidies will be included in the backpopulating'
+            ),
+        )
+
+    def process_transaction(self, subsidy, txn):
+        """
+        Given a transaction (and it's subsidy), backpopulate the email and title
+        """
+        logger.info(f"Processing {subsidy.uuid} transaction {txn.uuid}")
+
+        if txn.lms_user_email is None:
+            lms_user_email = subsidy.email_for_learner(txn.lms_user_id)
+            txn.lms_user_email = lms_user_email
+            logger.info(f"Found {lms_user_email} for {subsidy.uuid} transaction {txn.uuid}")
+        if txn.content_title is None:
+            content_title = subsidy.title_for_content(txn.content_key)
+            txn.content_title = content_title
+            logger.info(f"Found {content_title} for {subsidy.uuid} transaction {txn.uuid}")
+
+        if not self.dry_run:
+            txn.save()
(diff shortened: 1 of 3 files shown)


enterprise_subsidy/apps/transaction/management/commands/backpopulate_transaction_email_and_title.py
```diff
@@ -0,0 +1,65 @@
...
+        """
+        parser.add_argument(
+            '--dry-run',
+            action='store_true',
+            dest='dry_run',
+            help=(
+                'If set, no updates will occur; will instead log '
+                'the actions that would have been taken.'
+            ),
+        )
```
iloveagent57, line 39: Consider whether we want to optionally skip internal-only subsidies (or maybe skip those by default).

enterprise_subsidy/apps/transaction/management/commands/backpopulate_transaction_email_and_title.py
```diff
@@ -0,0 +1,65 @@
...
+    def handle(self, *args, **options):
+        """
+        Send it
+        """
+        if options.get('dry_run'):
+            self.dry_run = True
+            logger.info("Running in dry-run mode. No updates will occur.")
+
+        incomplete_only_filter = Q(lms_user_email__isnull=True) | Q(content_title__isnull=True)
+        for items_batch in batch_by_pk(Transaction, extra_filter=incomplete_only_filter):
```
iloveagent57, line 50: Nit: `batch_by_pk` is really cool. OTOH, you could loop over Subsidies first to do some level of batching for you, and then the sequence of work would have some level of predictability, and you'd only need to fetch each subsidy record once.  It also opens the door for us to do bulk fetches (filtered by the related subsidy record's enterprise uuid) in the future.

enterprise_subsidy/apps/transaction/management/commands/backpopulate_transaction_email_and_title.py
johnnagro: tell me more?

iloveagent57: We might have test subsidy records for say, performance testing, where there's lots and lots of test transaction records.  We might want the ability to have this command ignore test subsidy records, so we don't have to wait on those being updated.

enterprise_subsidy/apps/transaction/management/commands/backpopulate_transaction_email_and_title.py
```diff
@@ -0,0 +1,76 @@
...
+        )
+
+    def process_transaction(self, subsidy, txn):
+        """
+        Given a transaction (and it's subsidy), backpopulate the email and title
+        """
+        logger.info(f"Processing {subsidy.uuid} transaction {txn.uuid}")
+
+        if txn.lms_user_email is None:
+            lms_user_email = subsidy.email_for_learner(txn.lms_user_id)
```
iloveagent57, line 48: Do we want to make these calls if `self.dry_run` is true?  I could go either way - on the one hand, it's nice to set `--dry-run` _just_ to see what would be updated, and then have that happen relatively quickly.  On the other hand, it's nice to know with a dry run if we're actually fetching the data we expect.

enterprise_subsidy/apps/transaction/management/commands/backpopulate_transaction_email_and_title.py
```diff
@@ -0,0 +1,76 @@
...
+
+        for subsidies in batch_by_pk(Subsidy):
+            for subsidy in subsidies:
+                logger.info(f"Processing subsidy {subsidy.uuid}")
+                subsidy_filter = Q(ledger=subsidy.ledger)
+                incomplete_only_filter = Q(lms_user_email__isnull=True) | Q(content_title__isnull=True)
+                txn_filter = subsidy_filter & incomplete_only_filter
+                for txns in batch_by_pk(Transaction, extra_filter=txn_filter):
+                    for txn in txns:
+                        self.process_transaction(subsidy, txn)
```
iloveagent57, line 76: Love this refactoring

enterprise_subsidy/apps/transaction/management/commands/backpopulate_transaction_email_and_title.py
```diff
@@ -0,0 +1,76 @@
...
+    def process_transaction(self, subsidy, txn):
+        """
+        Given a transaction (and it's subsidy), backpopulate the email and title
+        """
+        logger.info(f"Processing {subsidy.uuid} transaction {txn.uuid}")
+
+        if txn.lms_user_email is None:
+            lms_user_email = subsidy.email_for_learner(txn.lms_user_id)
+            txn.lms_user_email = lms_user_email
+            logger.info(f"Found {lms_user_email} for transaction {txn.uuid}")
```
iloveagent57, line 50: nit: log the subsidy uuid in these logs, too

enterprise_subsidy/apps/transaction/management/commands/backpopulate_transaction_email_and_title.py
johnnagro: i think we wanna be able to see that these API calls work without actually saving the results

iloveagent57 (approved):

PR #185, opened 2023-11-27
feat: allow redemption at requested prices

iloveagent57:
Assignments (in enterprise-access) contain a field that represents the price of the content at the time of assignment. We need to allow for that price to be passed through to the transaction creation view. ENT-8035

See updated API docs at http://localhost:18280/api/schema/redoc/#tag/transactions/operation/api_v2_subsidies_admin_transactions_create

Use curl/postman to make a request to create a new transaction for a price that's slightly different from the true price, e.g.
```
curl --location 'http://localhost:18280/api/v2/subsidies/4c53f616-5ea0-42a0-9430-ba165cb2c916/admin/transactions/' \
--header 'Authorization: JWT [your-jwt]' \
--header 'Content-Type: application/json' \
--data '{
    "subsidy_access_policy_uuid": "[your policy uuid]",
    "lms_user_id": [your user id int],
    "content_key": "edX+DemoX",
    "requested_price_cents": 14200
}'
```

### Merge checklist
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Commits are squashed


enterprise_subsidy/apps/subsidy/models.py
```diff
@@ -499,7 +517,7 @@ def _create_redemption(
                 All other exceptions raised during the creation of an enrollment.  This should have already triggered
                 the rollback of a pending transaction.
         """
-        quantity = -1 * self.price_for_content(content_key)
+        quantity = -1 * content_price
```
iloveagent57, line 520: IDK why we made a second call to get the price here - we already have it from calling `is_redeemable()` further up in the body of `redeem()`.

pwnage101 (approved):
LGTM!

PR #190, opened 2023-12-21
ENT-8079 - Make DSC optional based on enterprise config

zwidekalanga:
### Description

### Testing instructions

Add some, if applicable

### Merge checklist
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Commits are squashed

### DIFF
--- a/enterprise_subsidy/apps/fulfillment/api.py
+++ b/enterprise_subsidy/apps/fulfillment/api.py
@@ -2,6 +2,8 @@
 Python API for interacting with fulfillment operations
 related to subsidy redemptions.
 """
+import logging
+
 from django.conf import settings
 from getsmarter_api_clients.geag import GetSmarterEnterpriseApiClient
 from openedx_ledger.models import ExternalFulfillmentProvider, ExternalTransactionReference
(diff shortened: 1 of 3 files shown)


enterprise_subsidy/apps/fulfillment/api.py
```diff
@@ -121,7 +151,11 @@ def _validate(self, transaction):
 
         Raises an exception when the transaction is missing required information
         """
+        enterprise_customer_data = self._get_enterprise_customer_data(transaction)
+        enable_data_sharing_consent = enterprise_customer_data.get('enable_data_sharing_consent', False)
         for field in self.REQUIRED_METADATA_FIELDS:
```
adamstankiewicz, line 156: [curious] I'm wondering if it'd make sense to conditionally push `geag_data_share_consent` onto `self.REQUIRED_METADATA_FIELDS` when `enable_data_sharing_consent = True`, so that the fields listed in `self.REQUIRED_METADATA_FIELDS` are truly only the required fields?

adamstankiewicz: Got it. I suppose the other concern is that I'm not sure we'd be able to get at the appropriate enterprise customer within a class constructor, as we are relying on a specific transaction to identify the enterprise customer. I think how you have it is reasonable.

adamstankiewicz (approved):

PR #203, opened 2024-02-08
feat: adding a non-customer dependent content metadata api client

alex-sheehan-edx:
### Description
Right now all of our enterprise catalog content metadata API client methods are depending on customer because the catalog service only surfaced content through a customer's context. This was too rigid for our purposes in the transaction reversal procedure where we needed to query for the course start date regardless of whether or not the content belonged under the customer's catalog. I've updated the catalog service to surface content metadata records without tying the serializer to the customer, so that the subsidy service can query and find the information it needs without having the association between content and customer catalog.

### Merge checklist
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Commits are squashed

### DIFF
--- a/enterprise_subsidy/apps/api_client/tests/test_enterprise_catalog.py
+++ b/enterprise_subsidy/apps/api_client/tests/test_enterprise_catalog.py
@@ -83,10 +83,10 @@ class EnterpriseCatalogApiClientTests(TestCase):
         }
 
     @mock.patch('enterprise_subsidy.apps.api_client.base_oauth.OAuthAPIClient', return_value=mock.MagicMock())
-    def test_successful_fetch_course_content_metadata(self, mock_oauth_client):
+    def test_successful_fetch_course_content_metadata_for_customer(self, mock_oauth_client):
         """
-        Test the enterprise catalog client's ability to handle api requests to fetch content metadata from the catalog
-        service.
+        Test the enterprise catalog client's ability to handle api requests to fetch customer specific content metadata
+        from the catalog service.
         """
         mock_oauth_client.return_value.get.return_value = MockResponse(self.course_metadata, 200)
         enterprise_catalog_client = EnterpriseCatalogApiClient()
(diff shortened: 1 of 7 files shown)


enterprise_subsidy/apps/api_client/tests/test_enterprise_catalog.py
```diff
@@ -83,14 +83,27 @@ def setUpTestData(cls):
...
+        """
+        Test the enterprise catalog client's ability to handle api requests to fetch content metadata from the catalog
+        service.
+        """
+        mock_oauth_client.return_value.get.return_value = MockResponse(self.course_metadata, 200)
+        enterprise_catalog_client = EnterpriseCatalogApiClient()
+        response = enterprise_catalog_client.get_content_metadata(self.course_key)
+        assert response == self.course_metadata
+        assert mock_oauth_client.return_value.get.call_args.args[0] == enterprise_catalog_client.metadata_endpoint
+        assert mock_oauth_client.return_value.get.call_args.kwargs == {'params': {'content_keys': ['edX+DemoX']}}
```
iloveagent57, line 109: One good thing to do here is to call `response = enterprise_catalog_client.get_content_metadata(self.course_key)` a second time to exercise the cache, and assert that the oauth client is still only called once.

alex-sheehan-edx: not a bad idea

enterprise_subsidy/apps/content_metadata/tests/test_api.py
```diff
@@ -241,20 +241,37 @@ def test_tiered_caching_works(self, mock_catalog_client):
...
+
+        _ = ContentMetadataApi.get_content_metadata(self.course_key)
+
+        self.assertTrue(TieredCache.get_cached_response(cache_key).is_found)
+        self.assertEqual(
+            TieredCache.get_cached_response(cache_key).value,
+            {'the': 'metadata'},
+        )
+        self.assertEqual(
+            ContentMetadataApi.get_content_metadata(self.course_key),
```
iloveagent57, line 274: Cool, this works for exercising the cache!

enterprise_subsidy/apps/api_client/tests/test_enterprise_catalog.py
alex-sheehan-edx: [inform] the caching happens at the metadata api level, so subsequent calls to the catalog api client should make multiple calls - I updated the content_metadata api caching test to ensure there's only one enterprise catalog client `get_content_metadata' call

alex-sheehan-edx: https://github.com/openedx/enterprise-subsidy/pull/203/files#diff-2c4756b0ae94b7023c130ba2f022f68ab167b4d1400155988f3ab24b3de3c91dR277

iloveagent57 (approved):
One last nit, looks good!

PR #209, opened 2024-02-16
feat: supply `parent_content_key` to transaction when redeeming content

pwnage101:
By storing the parent_content_key (i.e. course key) in addition to the content_key (i.e. course run key) on the transaction object, downstream consumers of transactions can conveniently determine the course key without needing to either A) construct it themselves from the course run key (not future proof) or B) make an API request to fetch the course key (not performant when working with many transactions).

In support of ENT-8389 and ADR 0008 in openedx-ledger

Depends on: https://github.com/openedx/openedx-ledger/pull/60


iloveagent57 (commented):
This looks generally correct to me!

iloveagent57 (approved):

PR #221, opened 2024-03-20
fix: correct variant is included in summary data for content

pwnage101:
**Background**: There's an internal helper method to generate "summary" data for a piece of content, which is used all over this codebase to extract relevant bits of metadata from said content. The summary data is also used verbatim in the enterprise-subsidy content metadata API endpoint (`/api/v1/content-metadata/<content-identifier>/`). The problem was that this helper method relied on `additional_metadata` from the content metadata, but that dict is now deprecated.  Furthermore, when a course RUN key was passed to this helper method, the result would incorrectly include the variant_id of the advertised run, potentially a different run than request.

**This PR**: Change logic to actually fetch the `variant_id` of the requested course run, and fallback to that of the advertised course run.  Furthermore, it cuts `additional_metadata` out of the loop entirely.

ENT-8520

### DIFF
--- a/enterprise_subsidy/apps/content_metadata/api.py
+++ b/enterprise_subsidy/apps/content_metadata/api.py
@@ -97,14 +97,24 @@ class ContentMetadataApi:
                 return source_name
         return ProductSources.EDX.value
 
-    def get_geag_variant_id_for_content(self, content_data):
+    def get_geag_variant_id_for_content(self, content_identifier, content_data):
         """
-        Returns the GEAG ``variant_id`` or ``None``, given a dict of ``content_data``.
-        In the GEAG system a ``variant_id`` is aka a ``product_id``.
+        Get a GEAG ``variant_id`` for the given content.
+
+        Args:
+            content_identifier (str): The content for which to get a variant.
+            content_data (dict): The content metadata dict fetched from the enterprise-catalog API.
+
+        Returns:
+            bool: The ``variant_id`` if one is found, or None otherwise.
+
+        Notes:
+            * In the GEAG system a ``variant_id`` is aka a ``product_id``.
+            * The ``content_identifier`` may be a course run key, in which case the variant returned will correspond to
+              that specific run, instead of the advertised run.
         """
-        variant_id = None
-        if additional_metadata := content_data.get('additional_metadata'):
-            variant_id = additional_metadata.get('variant_id')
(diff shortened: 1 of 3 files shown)


enterprise_subsidy/apps/content_metadata/api.py
```diff
@@ -97,14 +97,24 @@ def product_source_for_content(self, content_data):
...
+        Notes:
+            * In the GEAG system a ``variant_id`` is aka a ``product_id``.
+            * The ``content_identifier`` may be a course run key, in which case the variant returned will correspond to
+              that specific run, instead of the advertised run.
         """
-        variant_id = None
-        if additional_metadata := content_data.get('additional_metadata'):
-            variant_id = additional_metadata.get('variant_id')
+        course_run_content = self.get_course_run(content_identifier, content_data)
+        variant_id = course_run_content.get('variant_id')
```
macdiesel, line 117: I'm thinking we should provide a default here to code a bit more defensively? `None` was a valid return from this method before, perhaps we should allow it still? 

I think I also have a question here about how the API should handle a dict miss here, as in why wasn't this data available in the first place but I don't have the wider view right now. Something to consider. Perhaps I'm thinking a step too far.

pwnage101: > I'm thinking we should provide a default here to code a bit more defensively? None was a valid return from this method before, perhaps we should allow it still?

Line 117 should still result in a `None` return value if no variant_id is found, right?

> I think I also have a question here about how the API should handle a dict miss here, as in why wasn't this data available in the first place but I don't have the wider view right now. Something to consider. Perhaps I'm thinking a step too far.

Hm, so given that `variant_id` under course run metadata is a relatively recent addition, I'd say we or the open source community are slightly liable to have them missing. But more common would be this function called on OCM content, in which case the variant_id will always be absent.  Point is, I think it's actually quite normal for for the key to be there and not be there, depending on the content type.

macdiesel: I re-read the docs and I stand corrected: 
> If default is not given, it defaults to None, so that this method never raises a [KeyError](https://docs.python.org/2/library/exceptions.html#exceptions.KeyError).

macdiesel (approved):

iloveagent57 (approved):
Looks good! Smallest possible suggestion - add a test assertion somewhere about the variant_id being null.

PR #230, opened 2024-04-11
feat: add fulfillment logging, clean up misc logging

iloveagent57:
1. Adds logging for when we successfully fulfill or un-fulfill a transaction.
2. Cleans up some extraneous logging.
3. Converts some logging from `info` to `error` (because those calls are in an error condition).

### DIFF
--- a/enterprise_subsidy/apps/content_metadata/api.py
+++ b/enterprise_subsidy/apps/content_metadata/api.py
@@ -246,9 +246,8 @@ class ContentMetadataApi:
                 course_details,
                 django_cache_timeout=CONTENT_METADATA_CACHE_TIMEOUT,
             )
-        logger.info(
-            f'Fetched course details {course_details} for content_identifier {content_identifier}',
-        )
+        else:
+            logger.warning('Could not fetch metadata for content %s', content_identifier)
         return course_details
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


enterprise_subsidy/apps/content_metadata/api.py
```diff
@@ -246,9 +246,8 @@ def get_content_metadata(content_identifier, **kwargs):
                 course_details,
                 django_cache_timeout=CONTENT_METADATA_CACHE_TIMEOUT,
             )
-        logger.info(
-            f'Fetched course details {course_details} for content_identifier {content_identifier}',
-        )
```
iloveagent57, line 251: We log this several hundred times a day, but it's the _entire_ content metadata blob each time.
Instead, let's log only when we couldn't fetch a given identifier.

enterprise_subsidy/apps/content_metadata/api.py
```diff
@@ -273,10 +272,10 @@ def get_content_metadata_for_customer(enterprise_customer_uuid, content_identifi
...
-            course_details,
-            enterprise_customer_uuid,
-            content_identifier,
-        )
+        else:
+            logger.warning(
+                'Could not fetch metadata for customer %s, content %s',
+                enterprise_customer_uuid,
+                content_identifier,
+            )
```
iloveagent57, line 280: same idea here

enterprise_subsidy/apps/fulfillment/api.py
```diff
@@ -99,11 +99,6 @@ def _get_enterprise_customer_data(self, transaction):
         # Check if data is already cached
         cached_response = request_cache(namespace=REQUEST_CACHE_NAMESPACE).get_cached_response(cache_key)
         if cached_response.is_found:
-            logger.info(
-                'subsidy_record cache hit '
-                f'enterprise_customer_uuid={self._get_enterprise_customer_uuid(transaction)}, '
-                f'subsidy_uuid={transaction.uuid}'
-            )
```
iloveagent57, line 106: Our caching is quite stable now, we don't need this for debugging purposes any more.

brobro10000 (approved):

PR #236, opened 2024-04-22
feat: adding api subsidy endpoint to surface aggregated subsidy enrollment data

alex-sheehan-edx:
https://2u-internal.atlassian.net/browse/ENT-8562

first of two steps. First expose aggregated enrollment data pertaining to particular subsidies from enterprise-subsidy via a new API endpoint. Then have access query and cache said new endpoint, requesting pages of enterprise group member data and zipping the data together to form a hydrated enterprise group members view.

### Testing instructions

run unit tests, alternatively hit `http://localhost:18280/api/v1/subsidies/<subsidy-UUID>/aggregated-enrollments` after fetching and running locally

### Merge checklist
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Commits are squashed

### DIFF
--- a/enterprise_subsidy/apps/api/v1/serializers.py
+++ b/enterprise_subsidy/apps/api/v1/serializers.py
@@ -307,6 +307,30 @@ class TransactionCreationRequestSerializer(serializers.ModelSerializer):
 
 
 # pylint: disable=abstract-method
+class SubsidyLearnerAggregateSerializer(serializers.Serializer):
+    """
+    Serializer for providing responses to queries about aggregate enrollment data for a particular subsidy
+    """
+    lms_user_id = serializers.IntegerField(
+        help_text='User ID associated with user tied to a subsidy\'s transaction.'
+    )
+    total = serializers.IntegerField(
+        help_text='Number of enrollments found for the user.'
+    )
+
+
+# pylint: disable=abstract-method
+class SubsidyLearnerAggregateRequestSerializer(serializers.Serializer):
+    """
+    Serializer for creating a subsidy request
+    """
+    subsidy_access_policy_uuid = serializers.UUIDField(
+        required=False,
+        help_text="UUID of the enterprise subsidy policy. Used to filter down aggregate subsidy enrollment data",
+    )
+
+
+# pylint: disable=abstract-method
 class CanRedeemResponseSerializer(serializers.Serializer):
     """
     Serializer for providing responses to queries about redeemability
--- a/enterprise_subsidy/apps/api/v1/views/subsidy.py
+++ b/enterprise_subsidy/apps/api/v1/views/subsidy.py
@@ -13,6 +13,7 @@ from rest_framework import exceptions, mixins, permissions, status, viewsets
 from rest_framework.authentication import SessionAuthentication
 from rest_framework.decorators import action
 from rest_framework.response import Response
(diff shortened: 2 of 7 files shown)


enterprise_subsidy/apps/api/v1/serializers.py
```diff
@@ -306,6 +306,30 @@ def create(self, validated_data):
...
+class SubsidyAggregateEnrollmentsRequestSerializer(serializers.Serializer):
+    """
+    Serializer for creating a subsidy request
+    """
+    subsidy_access_policy_uuid = serializers.UUIDField(
+        required=False,
+        help_text="UUID of the enterprise subsidy policy. Used to filter down aggregate subsidy enrollment data",
+    )
+
+
```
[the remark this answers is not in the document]
alex-sheehan-edx: oooh good suggestion!

enterprise_subsidy/apps/api/v1/serializers.py
```diff
@@ -306,6 +306,30 @@ def create(self, validated_data):
         return transaction
 
 
+# pylint: disable=abstract-method
+class SubsidyLearnerAggregateSerializer(serializers.Serializer):
```
kiram15, line 310: [question] Why are we using `serializers.Serializer` vs `serializers.ModelSerializer` like the examples above?

enterprise_subsidy/apps/api/v1/views/subsidy.py
```diff
@@ -371,3 +374,31 @@ def partial_update(self, request, *args, **kwargs):
...
         kwargs['partial'] = True
         return self.update(request, *args, **kwargs)
+
+    @extend_schema(
+        tags=['subsidy'],
+        request=SubsidyLearnerAggregateRequestSerializer,
+        parameters=[Parameters.ENTERPRISE_SUBSIDY_ACCESS_POLICY_UUID],
+        responses={
+            200: SubsidyLearnerAggregateSerializer,
+            403: exceptions.PermissionDenied,
```
kiram15, line 384: [nit] These should be in numerical order

enterprise_subsidy/apps/api/v1/serializers.py
iloveagent57: `ModelSerializer` is a convenience class for more easily serializing a Django model.  Since there's no model related to this flow (computing aggregates by learner), we can't use `ModelSerializer`.
https://www.django-rest-framework.org/api-guide/serializers/#modelserializer

iloveagent57 (commented):
Sorry, one more nit - you might want to make clear in the docs for this endpoint (just in the docstring is fine) that the response is not paginated, because the other endpoints under this view set _are_ paginated.

enterprise_subsidy/apps/api/v1/serializers.py
alex-sheehan-edx: +1 to what @iloveagent57 has to say- while we query the subsidy model, we're not directly returning the records but instead modifying and calculating an aggregate from the data. Not returning the records from the model means that a model serializer wouldn't work for us

enterprise_subsidy/apps/api/v1/views/subsidy.py
alex-sheehan-edx: oops hahaha nice catch

PR #258, opened 2024-06-24
feat: rollback_transaction() also cancels external fulfillments

pwnage101:
This is needed for cases where an external fulfillment has been created, but the subsequent platform fulfillment fails (e.g. due to transient networking issues).  Before this PR, the external fulfillment would not get rolled back, resulting in an "orphaned" fulfillment that allows learners to effectively redeem content for free.  After this PR, that loophole is fixed.

ENT-8866


enterprise_subsidy/apps/subsidy/models.py
```diff
@@ -417,13 +417,45 @@ def commit_transaction(self, ledger_transaction, fulfillment_identifier=None, ex
...
-        logger.info(f'Setting transaction {ledger_transaction.uuid} state to failed.')
-        ledger_transaction.state = TransactionStateChoices.FAILED
-        ledger_transaction.save()
+        try:
+            if external_transaction_reference and external_transaction_reference.transaction == ledger_transaction:
+                logger.info(
+                    '[rollback_transaction] Attempting to cancel external fulfillment for transaction %s.',
+                    ledger_transaction.uuid,
+                )
+                self.geag_fulfillment_handler().cancel_fulfillment(external_transaction_reference)
```
iloveagent57, line 440: Maybe one more `log.info()` after `cancel_fulfillment()` to declare success?

iloveagent57 (approved):
Very good, one logging suggestion.

PR #269, opened 2024-07-09
feat: Add data migration to backfill legacy initial deposits

pwnage101:
This needs to be located in the Subsidy app instead of openedx_ledger because it relies on having access to the SubsidyReferenceChoices class.

ENT-9075

BLOCKED on https://github.com/openedx/openedx-ledger/pull/96
BLOCKED on https://github.com/openedx/openedx-ledger/pull/97

### DIFF
--- a/enterprise_subsidy/apps/subsidy/migrations/0022_backfill_initial_deposits.py
+++ b/enterprise_subsidy/apps/subsidy/migrations/0022_backfill_initial_deposits.py
@@ -1,0 +1,83 @@
+"""
+Backfill initial deposits.
+
+Necessarily, this also backfills SalesContractReferenceProvider objects based on the values currently defined via
+SubsidyReferenceChoices.
+
+Note this has no reverse migration logic. Attempts to rollback the deployment which includes this PR will not delete
+(un-backfill) the deposits created during the forward migration.
+"""
+import django.utils.timezone
+from django.db import migrations
+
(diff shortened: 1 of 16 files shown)


enterprise_subsidy/apps/subsidy/migrations/0022_backfill_initial_deposits.py
```diff
@@ -0,0 +1,78 @@
...
+            "ledger": tx.ledger,
+            "transaction": tx,
+            "desired_deposit_quantity": tx.quantity,
+            "sales_contract_reference_id": tx.ledger.subsidy.reference_id,
+            "sales_contract_reference_provider": sales_contract_reference_providers[tx.ledger.subsidy.reference_type],
+        }
+        deposit = Deposit(**deposit_fields)
+        historical_deposit = HistoricalDeposit(
+            uuid=deposit.uuid,
+            history_date=datetime.now()
```
iloveagent57, line 57: We should use Django's TZ-aware `now()` here: https://docs.djangoproject.com/en/5.0/ref/utils/#django.utils.timezone.now

iloveagent57 (commented):
This is great, very clear and well-commented!

iloveagent57 (approved):

pwnage101:
I made one more change to add two more fields to the manually-created HistoricalDeposit:
```py
            history_type="+",
            history_change_reason="Data migration to backfill initial deposits",
```

PR #270, opened 2024-07-09
feat: add mgmt command to replay transaction reversal events

iloveagent57:
This will allow a consumer in enterprise-access to update related assignment record state for reversals that occurred before the producer/consumer logic was running to synchronize reversed state between ledger transactions and assignment records. ENT-9212

### Testing instructions
First, make sure you have some transactions with existing reversals.
Then make sure kafka is running, you can use `make dev.up.with-events` to do this.
Wait for kafka-control-center to be reachable.
In an enterprise-access container, run the consumer to read from the transaction lifecycle topic:
```
./manage.py consume_events -t enterprise-subsidies-transaction-lifecycle --group_id enterprise_access_dev
```

In an enterprise-subsidy container, replay the reversal events.
```
./manage.py replay_reversal_events
```

Observe in kafka that events are produced, and observe in enterprise-access logs that the consuming signal is processing them and updating state of any related assignment records.
Add some, if applicable

### Merge checklist
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Commits are squashed


enterprise_subsidy/apps/core/management/commands/produce_enterprise_ping_event.py
```diff
@@ -81,7 +81,6 @@ def add_arguments(self, parser):
     def handle(self, *args, **options):
         try:
             producer = create_producer()
-            # breakpoint()
```
pwnage101, line 84: delete

pwnage101 (approved):

enterprise_subsidy/apps/core/management/commands/produce_enterprise_ping_event.py
iloveagent57: I did! 😂

PR #290, opened 2024-08-20
fix: content_metadata enroll_by_date only drawn from course runs.

iloveagent57:
Since we *always* have a course run within the scope of `summary_data_for_content()` executions, the enroll by date should always be determined directly from a course run record, and not from the *course's* `normalized_metadata` dictionary (the latter of which will always contain the advertised run's enrollment deadline, regardless of the *requested* course run - we typically request course runs, and not courses, of this endpoint in prod). 
ENT-7472

### Description
The enterprise-subsidy CM endpoint calls the catalog customer CM endpoint.  If you pass the latter (catalog-customer-cm endpoint) a course run, you'll typically get a top-level _course_ back, with course runs inside of it. So despite typically receiving runs as input to the subsidy CM API, within the bounds of `summary_data_for_content()`, `content_data` is usually a course, not a run (because it's fetched from the catalog-customer-cm endpoint).

### Testing instructions

Add some, if applicable

### Merge checklist
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Commits are squashed

### DIFF
--- a/enterprise_subsidy/apps/content_metadata/api.py
+++ b/enterprise_subsidy/apps/content_metadata/api.py
@@ -124,29 +124,24 @@ class ContentMetadataApi:
                 variant_id = additional_metadata.get('variant_id')
         return variant_id
 
-    def enroll_by_date_for_content(self, content_data, content_mode):
+    def enroll_by_date_for_content(self, course_run_data, content_mode):
         """
-        Determines the enrollment deadline for the given ``content_data``,
-        which could be either a course or course run, hence the branching
-        on content type below.
+        Determines the enrollment deadline for the given ``course_run_data``.
 
         Args:
-            content_data: A dictionary of metadata pertaining to a course or course run.
-            content_mode: The (already computed) course mode for the content.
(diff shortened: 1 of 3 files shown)


enterprise_subsidy/apps/content_metadata/api.py
```diff
@@ -124,29 +124,24 @@ def get_geag_variant_id_for_content(self, content_identifier, content_data):
...
-            content_mode: The (already computed) course mode for the content.
+            course_run_data: A dictionary of metadata pertaining to a course run.
+            content_mode: The (already computed) course mode for the course run.
 
         Returns:
             A datetime string, or possibly null.
         """
-        if content_data.get('content_type') == 'course':
-            return content_data.get('normalized_metadata', {}).get('enroll_by_date')
-
```
pwnage101, line 142: I'm having trouble understanding how this changed any behavior if we only ever sent runs to this function, because in that case we'd never have run this conditional block, right?

pwnage101 (approved):
LGTM because it definitely looks correct, just don't quite understand how this actually changes behavior.

PR #316, opened 2024-11-13
feat: update get_content_metadata_for_customer to use v2 endpoint

pwnage101:
ENT-9410

### DIFF
--- a/enterprise_subsidy/apps/api_client/enterprise_catalog.py
+++ b/enterprise_subsidy/apps/api_client/enterprise_catalog.py
@@ -16,9 +16,13 @@ class EnterpriseCatalogApiClient(BaseOAuthClient):
     """
     API client for calls to the enterprise service.
     """
-    api_base_url = urljoin(settings.ENTERPRISE_CATALOG_URL, 'api/v1/')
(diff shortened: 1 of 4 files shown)


enterprise_subsidy/apps/api_client/enterprise_catalog.py
```diff
@@ -16,7 +16,7 @@ class EnterpriseCatalogApiClient(BaseOAuthClient):
     """
     API client for calls to the enterprise service.
     """
-    api_base_url = urljoin(settings.ENTERPRISE_CATALOG_URL, 'api/v1/')
+    api_base_url = urljoin(settings.ENTERPRISE_CATALOG_URL, 'api/v2/')
```
iloveagent57, line 19: Do we need every v1 endpoint in enterprise-catalog to have a functional v2 version (even if using the same exact logic as v1)?

pwnage101: Nope, I was thinking of elaborating on this by adding a constructor argument to ask for the API version.

iloveagent57 (approved):

PR #331, opened 2025-01-08
fix: add data object serialization support for reversal writing

pwnage101:
At some point during initial development I assumed the serialized data objects from openedx_events supported dict-like access patterns. I was wrong. This commit generalizes access patterns to support both dicts and objects so that we can continue to support both event-driven and cron-driven reversal writing.

### DIFF
--- a/enterprise_subsidy/apps/transaction/tests/test_signal_handlers.py
+++ b/enterprise_subsidy/apps/transaction/tests/test_signal_handlers.py
@@ -10,6 +10,11 @@ import ddt
 import pytest
 from django.test import TestCase
 from django.test.utils import override_settings
+from openedx_events.enterprise.data import (
+    EnterpriseCourseEnrollment,
+    EnterpriseCustomerUser,
+    LearnerCreditEnterpriseCourseEnrollment
+)
 from openedx_ledger.models import TransactionStateChoices
 from openedx_ledger.signals.signals import TRANSACTION_REVERSED
 from openedx_ledger.test_utils.factories import (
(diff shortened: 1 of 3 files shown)


enterprise_subsidy/apps/transaction/tests/test_signal_handlers.py
```diff
@@ -10,6 +10,11 @@
 import pytest
 from django.test import TestCase
 from django.test.utils import override_settings
+from openedx_events.enterprise.data import (
+    EnterpriseCourseEnrollment,
+    EnterpriseCustomerUser,
+    LearnerCreditEnterpriseCourseEnrollment
+)
```
iloveagent57, line 17: Good call.

enterprise_subsidy/apps/transaction/utils.py
```diff
@@ -60,14 +72,14 @@ def unenrollment_can_be_refunded(
 
     Args:
       content_metadata (dict): Metadata for course from which the learner has been unenrolled.
-      enterprise_course_enrollment: (dict-like):
+      enterprise_course_enrollment: (dict or openedx_events.enterprise.data.EnterpriseCourseEnrollment):
```
iloveagent57, line 75: What if this function did something like below to cast dicts into `EnterpriseCourseEnrollments`?
```python
record = enterprise_course_enrollment
if isinstance(record, dict):
    record = EnterpriseCourseEnrollment(**record)
# then use record in all the logic below.
```
Or better yet, maybe the _caller_ should be forced to do that casting, and this function can have a stricter contract, that it only accepts EnterpriseCourseEnrollments?

iloveagent57 (commented):
Seems legit, small suggestion on stuff in `utils.py`.

enterprise_subsidy/apps/transaction/utils.py
pwnage101: Yea, I briefly considered forcing the onus of conversion on the caller, but I wasn't sure the object could be coerced into a dict so simply because of the multiple nested layers of objects. However, now that I look at how `enterprise_course_enrollment` is actually used I think it's feasible to do it your way because only the top level keys are accessed by `unenrollment_can_be_refunded()`.

pwnage101: I had to coerce the other way around because the EnterpriseCourseEnrollments class would not accept the partial data which the `/unenrolled` endpoint returns. Adding `.__dict__` onto the end of an event data object seemed to work fine, for one level of data:

```
(Pdb) test_lc_course_enrollment.__dict__
{'uuid': UUID('07c02939-6a66-4006-9e0c-5ef3afcb05d9'), 'created': datetime.datetime(2020, 1, 1, 12, 0), 'modified': datetime.datetime(2020, 1, 1, 12, 0), 'fulfillment_type': 'learner_credit', 'is_revoked': True, 'enterprise_course_entitlement_uuid': None, 'enterprise_course_enrollment': EnterpriseCourseEnrollment(id=1, created=datetime.datetime(2020, 1, 1, 12, 0), modified=datetime.datetime(2020, 1, 1, 12, 0), enterprise_customer_user=EnterpriseCustomerUser(id=1, created=datetime.datetime(2020, 1, 1, 12, 0), modified=datetime.datetime(2020, 1, 1, 12, 0), enterprise_customer_uuid=UUID('2a4329a7-bc2e-4042-8c1e-e2e5cc335ae0'), user_id=1, active=True, linked=True, is_relinkable=True, should_inactivate_other_customers=True, invite_key=None), course_id='course-v1:bin+bar+baz', saved_for_later=False, source_slug=None, unenrolled=True, unenrolled_at=datetime.datetime(2020, 1, 1, 0, 0)), 'transaction_id': UUID('b9f7a49f-3898-40fa-85ab-82337957a3c5')}
```

iloveagent57 (approved):

PR #372, opened 2025-06-25
Add support for django 5.2

irtazaakram:
- resolves https://github.com/openedx/public-engineering/issues/376


awais786:
Just one change needed — the rest of the PR looks good. Please revert the migration that's showing up due to django-simple-history.

enterprise_subsidy/apps/subsidy/migrations/0023_alter_historicalsubsidy_options_and_more.py
```diff
@@ -0,0 +1,21 @@
+# Generated by Django 5.2.3 on 2025-06-25 07:42
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
```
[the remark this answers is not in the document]
feanil: Do we want to keep this, this seems fine since it's a  non schema change.

requirements/common_constraints.txt
```diff
@@ -25,7 +25,7 @@ Django<5.0
 elasticsearch<7.14.0
 
 # django-simple-history>3.0.0 adds indexing and causes a lot of migrations to be affected
-3.0.0
```
feanil, line 28: It seems like maybe this was mis-formatted?  Should the comment above still exist if this is being removed?

feanil: Ok, sounds good.  That's upstream weirdness and we can look at it over there.

awais786 (approved):

PR #377, opened 2025-07-21
fix: create deposit api view less restrictive

iloveagent57:
Allow for the creation of deposits as long as the related subsidy has not yet expired, even if the subsidy/ledger is not yet *active*.
ENT-10615

Testing:
Hit this endpoint on a subsidy that has already expired:
```
curl --location 'http://localhost:18280/api/v2/subsidies/4025047c-609a-4e64-8546-aaddb15af4eb/admin/deposits/' \
--header 'Authorization: JWT [your-jwt]
--header 'Content-Type: application/json' \
--data '{does not actually matter for this test}'
}'
```

### Merge checklist
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Commits are squashed

### DIFF
--- a/enterprise_subsidy/apps/api/v2/tests/test_deposit_views.py
+++ b/enterprise_subsidy/apps/api/v2/tests/test_deposit_views.py
@@ -2,8 +2,10 @@
 Tests for the v2 deposit views.
 """
 import uuid
+from datetime import timedelta
 
 import ddt
+from django.utils import timezone
 from openedx_ledger.models import Deposit, SalesContractReferenceProvider
 from rest_framework import status
 from rest_framework.reverse import reverse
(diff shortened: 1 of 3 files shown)


enterprise_subsidy/apps/api/v2/tests/test_deposit_views.py
```diff
@@ -113,9 +115,12 @@ def test_deposit_creation(
...
         self.set_up_operator()
 
-        subsidy = SubsidyFactory(enterprise_customer_uuid=STATIC_ENTERPRISE_UUID)
+        subsidy = SubsidyFactory(
+            enterprise_customer_uuid=STATIC_ENTERPRISE_UUID,
+            active_datetime=timezone.now() + timedelta(days=1),
+        )
         if not subsidy_active:
-            subsidy.expiration_datetime = subsidy.active_datetime
+            subsidy.expiration_datetime = timezone.now() - timedelta(days=1)
```
pwnage101, line 123: I think cases with `subsidy_active=False` causes the test subsidy to have an `expiration_datetime` _before_ `active_datetime`, which wouldn't happen in practice.  Furthermore, the condition name for executing this if-block is misleading because the subsidy will never be active in any test case anymore, given `active_datetime` will always be in the future.  I think what you want is something more like this:

```python
if subsidy_active:
    active_datetime = timezone.now() - timedelta(days=1)
    expiration_datetime = timezone.now() + timedelta(days=1)
else:
    if subsidy_expired:
        active_datetime = timezone.now() - timedelta(days=2),
        expiration_datetime = timezone.now() - timedelta(days=1)
    else:
        active_datetime = timezone.now() + timedelta(days=1),
        expiration_datetime = timezone.now() + timedelta(days=2)    
```

This also would introduce `subsidy_expired` which helps explicitly cover both when the inactive subsidy just hasn't started yet vs. has already expired.

pwnage101 (approved):
Perfect!

PR #389, opened 2025-09-04
feat: forced enrollment and client-side GEAG idempotency for "duplicate order" error

pwnage101:
In order to mitigate the "duplicate order" error from the GEAG create enterprise allocation endpoint, we should try a little harder to search for an Order UUID / external reference ID to attach to the transaction.

First, do not instantly give up just because we recieved a "duplicate order" error. Then search for the Order UUID in possibly two places:

1. On the error response payload. The ID is unlikely to be here today, but it might in the future.
2. On the transaction metadata. The ID is also unlikely to be here in general, but can be populated when utilizing ForcedPolicyRedemption.

ENT-10879


enterprise_subsidy/apps/fulfillment/api.py
```diff
@@ -15,9 +15,10 @@
 from enterprise_subsidy.apps.core.utils import request_cache, versioned_cache_key
 from enterprise_subsidy.apps.subsidy.constants import CENTS_PER_DOLLAR
 
-from .constants import EXEC_ED_2U_COURSE_TYPES, OPEN_COURSES_COURSE_TYPES
+from .constants import EXEC_ED_2U_COURSE_TYPES, FALLBACK_EXTERNAL_REFERENCE_ID_KEY, OPEN_COURSES_COURSE_TYPES
 from .exceptions import FulfillmentException, InvalidFulfillmentMetadataException
 
+GEAG_DUPLICATE_ORDER_ERROR_CODE = 10174
```
iloveagent57, line 21: tiny nit: could be stored in settings for flexibility.

enterprise_subsidy/apps/fulfillment/tests/test_api.py
```diff
@@ -259,24 +263,76 @@ def test_validate_fail(self, mock_get_enterprise_customer_data):
 
     @mock.patch("enterprise_subsidy.apps.content_metadata.api.ContentMetadataApi.get_content_summary")
     @mock.patch("enterprise_subsidy.apps.api_client.enterprise.EnterpriseApiClient.get_enterprise_customer_data")
+    @ddt.data(
```
iloveagent57, line 266: nice

iloveagent57 (approved):
very nice

PR #395, opened 2025-09-17
fix: check if we're doing a geag fulfillment *and* have a variant id

iloveagent57:
### Description
Fixes a case where:
1. we're redeeming an exec ed transaction.
2. there's no variant id available for the content we're redeeming

This resulted in us gracefully skipping over external fulfillment, but still continuing with the edx enrollment and the transaction being committed.

### Testing instructions

I've only tested via unit tests so far.

### Merge checklist
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Commits are squashed

### DIFF
--- a/enterprise_subsidy/apps/subsidy/models.py
+++ b/enterprise_subsidy/apps/subsidy/models.py
@@ -32,7 +32,8 @@ from enterprise_subsidy.apps.api_client.lms_user import LmsUserApiClient
 from enterprise_subsidy.apps.content_metadata.api import ContentMetadataApi
 from enterprise_subsidy.apps.core import event_bus
 from enterprise_subsidy.apps.core.utils import localized_utcnow
-from enterprise_subsidy.apps.fulfillment.api import GEAGFulfillmentHandler
+from enterprise_subsidy.apps.fulfillment.api import GEAGFulfillmentHandler, is_geag_fulfillment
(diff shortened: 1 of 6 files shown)


enterprise_subsidy/apps/subsidy/models.py
```diff
@@ -628,15 +629,25 @@ def _create_redemption(
...
+        if is_geag_fulfillment(ledger_transaction):
+            if not self.geag_fulfillment_handler().can_fulfill(ledger_transaction):
+                logger.error(
+                    f'Could not find variant id for transaction {ledger_transaction.uuid} '
+                    f'which can only be fulfilled via GEAG. Rolling back transaction.'
+                )
+                self.rollback_transaction(ledger_transaction)
+                raise IncompleteContentMetadataException(
+                    f'Missing variant_id needed for GEAG transaction {ledger_transaction}, not attempting fulfillment'
+                )
```
pwnage101, line 641: Nit: I'd prefer we raise and just let the exception handler perform rollback so that we funnel all rollback scenarios to one line.

pwnage101 (approved):

enterprise_subsidy/apps/subsidy/models.py
iloveagent57: ah, yup, that's cleaner

